PR #11, opened 2016-09-28
JSON Updates in .XML files under src

Kritner:
not sure if this is needed/wanted, but while going through the spec I noticed some issues with some of the JSON.
- There are likely more files exhibiting inconsistencies that I have not yet read through
- I ***did not*** run the MakeFile, as it was making more changes to the HTML/TXT files under artifacts than the changes I made would have warranted. I'm assuming the XML files are the "source" and the HTML/TXT files are intended to be generated from that XML, however it seems that they have not been kept in sync within the repo.


Kritner:
@bfussell thanks for the approval, was there anything additional you needed before this is merged into the repo?  

I'm assuming you saw I only updated the source, and not the generated HTML/XML.  Is this something you can do?  If not, I could run the MakeFile, and append those changes to the PR.  I was just worried about adding that into the initial PR, as it seemed like that would overwrite some things that were done only to the HTML or TXT files.

Issue #7, opened 2016-09-19
Need hooks for data expiration policy and management in ACVP


atvassilev:
A new issue from NIST Sprint 1:

Need to establish some boundaries around when data will expire. This includes:
How long should registration/capabilities be available before test vectors are retrieved?
How long should test vectors be available after they are retrieved? It is it even necessary to keep them?
How long should a test id be active before it is no longer possible to submit test results?
Should the length of time be based on last interaction or total time taken?

Need to think how all these should be handled by the ACV protocol.

jfigus:
I'll take a look at this and draft a new section in the top-level spec to address expiration.

atvassilev:
I reviewed the text. It looks good, just a couple of suggestions for improvement.

1. Section 12, end of first paragraph" "... when the the client attempts to download the vector set or submit KATs for validation. 

2. In section 12.1 the implied behavior is that submitting results for validation does NOT extend the expiry date. In my view the text also implies that the client can submit as many corrected responses as needed as long as all of them arrive before the expiry date. This is fine with me, just suggest the following sentence be added to third paragraph in Section 12, right after the first sentence" "... The server may change the expiration timestamp of a previously issued vector set to either extend or shorten its lifetime subject to server policy." Here I am thinking about cases when the server discovers an error in the tests and wants to invalidate vector sets already sent to clients.

Please let me know if you have have questions or concerns.

jfigus:
Thanks for the feedback.  I'll add the suggested text that you provided.

atvassilev:
Fixed with revisions 15234a8 and 2ae0dfa

PR #16, opened 2016-09-30
PR - updates to artifacts, new windows batch file to generate artifacts on windows

Kritner:
- Created `WindowsGenerateArtifacts.bat` for use in recreating `artifacts` from `src` in a Windows environment.
- Updated all `artifacts` from `src`
- Updated `README.md` with basic instructions for keepings `artifacts` an `src` synchronized.


Kritner:
Looks like there are some issues in the artifacts that would conflict due to the other pull request.  Can probably resolve this by recreating the artifacts from the XML.  I can try to do that today.

atvassilev:
Just to double-check: earlier today I merged #7. If this pull request gets merged, it would not invalidate the updates from #7 in src but the artifacts will not be current since you generated them from the previous version of acvp_protocol.xml, correct?

Kritner:
@atvassilev in commit fc7f93f48774bbbea17300f98865636db255214e I regenerated the protocol artifacts  based on the new .xml file from #17.  This can be confirmed by searching "sub-document" as that was a typo corrected, as well as the new section "Vector Set Expiration".  My changes do not take into account 719987b63d2bb193ca1218d520a1d2a644d7e70d. I can update that now.

Kritner:
Updated artifacts with changes from 719987b63d2bb193ca1218d520a1d2a644d7e70d

Issue #14, opened 2016-09-30
Symmetric Model Section 3.2 Name Clarification


DBugger32:
Currently we have these sections...
3. Test Vectors
3.1 Test Groups JSON Schema
3.2 Test Vectors JSON Schema

I believe this is confusing as Test Vector is used twice in the hierachy .  Recommend we go to the following...
3. Test Vectors
3.1 Test Groups
3.2 Test Cases

and change all references to Test Vector in the Section 3.2 text to Test Case

> Each test group contains an array of one or more test cases. Each test case is a JSON object that represents a single test case to be processed by the ACVP client. The following table describes the JSON elements for each test case.

jfigus:
Good catch, I'll make this correction.

jfigus:
Fixed in 121b2c2faf5553f2a03b328f683260fc40105220

Issue #13, opened 2016-09-30
Symmetric Unique Numeric Test Id Scope


DBugger32:
In Section 3.2, tcId is defined as "Unique numeric identifier for the test case".  Is this id unique within the context of the Test Group or all the Test Cases within the Test Vector?

jfigus:
Unique among the entire vector set.  I'll update the spec to clarify this.

jfigus:
fixed in 121b2c2faf5553f2a03b328f683260fc40105220

PR #18, opened 2016-10-06
Clarify uniqueness requirement for test case ID. Eliminate overloaded…

jfigus:
This addresses issues #13 and #14.  If this change looks good and is accepted, I can update all the other sub-specs to use this new verbiage.

### DIFF
--- a/src/acvp_sub_symmetric.xml
+++ b/src/acvp_sub_symmetric.xml
@@ -371,10 +371,10 @@
 	    </texttable>
 	</section>
 
-	<section title="Test Vectors JSON Schema" anchor="tvjs">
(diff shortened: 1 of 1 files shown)


Kritner:
If at all possible, could #16 be merged prior to #18?  This PR does not contain updates to artifacts as the comment indicates (at least as of this comment), and #16 has updates to all artifacts.  If this PR (#18) is merged first, #16 will likely get merge conflicts again.

jfigus:
That's fine with me.  In the future, I'm thinking we should only take pull requests for the source files.  Then the artifacts would be regenerated and committed periodically to the repo by one of the repo owners.

atvassilev:
This would be fine with me.

jfigus:
OK, I'll submit another PR with the same change to the other sub-specs.

Issue #25, opened 2016-11-02
Sample JSON in section 11.2 is not consistent with symmetric cipher sub-spec


jfigus:
The "mode" value is no longer defined in the sub-spec.  The length names are no longer plural.

 "mode" : "both",
   "ivGen" : "internal",
   "ivGenMode" : "8.2.1",
   "keyLens" : [
               128,
               256
                ],
   "tagLens" : [
               96
                  ],
   "ivLens" : [
               96
                  ],
   "ptLens" : [
               0,
               128,
               136
                  ],
"aadLens" : [

atvassilev:
PR #31 and #33  resolve this issue.

Issue #26, opened 2016-11-22
Need failure reason in test results response from ACVP server


jfigus:
Do we want to add a reason failure to the test results in section 11.4 for failed test cases? The JSON response would look something like this...

   {
   "version" : "0.2",
   "results" : {
       "vsId" : 1437,
       "disposition" : "incomplete",
       "tests" : [
               { "tcId" : 12340, "result" : "passed", "reason" : "" },
               { "tcId" : 12341, "result" : "incomplete", "reason" : "" },
               { "tcId" : 12342, "result" : "failed", "reason" : "Authentication tag failure" }
                 ]
      }
}

atvassilev:
I am in favor.

bfussell:
Sounds good to me.  As we have discussed on a few occasions the error reporting is presently
pretty weak and we need to make sure we continue to enhance it.  I still have an action item
to review the CAVS tool and make sure we cover all errors it detected.

Barry

sam1689:
Good idea - I think it will enhance the value of the testing and save everyone more time if we can include a decent failure reason in the results.

-Sam

atvassilev:
PR #31 resolves this.

PR #31, opened 2016-12-22
Add files via upload

atvassilev:
This update resolves Issues #25 and #26. Since John is no longer with the Cisco team and is not known if and when he would be able to join the project again, I took care of the proposed changes, given there is broad agreement on them.

### DIFF
--- a/src/acvp_protocol.xml
+++ b/src/acvp_protocol.xml
@@ -47,8 +47,7 @@
 
    <!-- Another author who claims to be an editor -->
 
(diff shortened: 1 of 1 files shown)


atvassilev:
I see. I did not know about the lower case issue and missed oen 's' by mistake since my local file got corrupted and had to redo it from a clean copy. Since you merged #32, how about merging this one too and I will take care of the rest of the issues in a follow up PR? That way the src and artifacts will be in sync.

Issue #10, opened 2016-09-27
Protocol specification needs additional detail for vectors and results


hbooth:
Additional detail is needed in the high-level protocol specification to describe the vectors and results URLs. Obviously, the specific details for each algorithm are in the algorithm sub-specifications, but a high-level description of what the expected inputs are (pointing out the details are in the sub specs) and expected response is needed.

hbooth:
This looks to have been largely taken care of. If anything further crops up we can open a new issue.

PR #40, opened 2017-01-09
URI info

bfussell:
This is a start - if more detail is required please provide some comments.


hbooth:
I like the direction and had the following comments:

Looking at 11.3, It might be helpful to indicate that the contents of the test group will vary depending on the specific sub-specification. I know this is obvious to us, but I don't think it will hurt to be explicit.

I looked and could not find a description of the general process of the POST for providing the Vector Test Response. It should also get a mention in section 4.2. Maybe Table 1 should have four columns:
Operation, URL, HTTP Method, Details

where HTTP METHOD is POST or GET (currently) since the client and server will have to handle these differently for a given URL and some URLs support both POST and GET.

What do you think?

atvassilev:
I will leave the corresponding issue #10 open until after the rest of Harold's feedback is incorporated in the spec.

PR #43, opened 2017-01-24
Update acvp_protocol.xml

dmchodnicki:
KAT Exchange and other ACVP spec clarifications.

### DIFF
--- a/src/acvp_protocol.xml
+++ b/src/acvp_protocol.xml
@@ -234,7 +234,7 @@ The following functions are outside the scope of this document:</t>
 +-----------------------------------------------+
 | HTTP[S] message transfer and signaling        |
 +-----------------------------------------------+
-| TLS for transport security(optional)          |
+| TLS for transport security(recommended)       |
 +-----------------------------------------------+
 | TCP for transport                             |
 |_______________________________________________|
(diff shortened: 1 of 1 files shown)


Kritner:
Looks better this time :)

atvassilev:
I am ok with it but defer to @bfussell for the merge.

bfuss:
I don't have an issue moving the URI information into the Message Format section, however
it appears you started with a version of acvp_portocol.xml that did not include my last commit.
There is some text missing from 11.3 and an entire section entitiled  4.2 HTTP URI Operations that
is missing.

I'm not sure of the group's opinion, but I would also prefer the artifacts updated when making changes
to the .xml file.

dmchodnicki:
I used GitHub to pull in the latest committed version of the file before starting my edits so I thought I maybe made a cut and past error.  However, I did a thorough recheck this morning and didn't see any evidence of missing content in either 11.3 or 4.2.  In fact 4.2 HTTP URI Operations is the section where the table was expanded to include an extra column for HTTP Method and the extra row for Submit Vector Test Results.  Please let me know if I'm still missing something.  I can be reached at 910-833-8040 any time this morning.

Regarding the artifacts, the intent was to use the WindowsGenerateArtifacts.bat file to update the artifacts once the updated xml file was approved.  I brought this up in this morning's scrum and the team prefers this approach of one Pull Request on the src files instead of 2 separate Pull Requests (e.g., one for the src files and one for the corresponding artifacts).

Kritner:
Hey @bfuss I don't know if this could be where some of the confusion is coming from or not, but the default "comparison" view collapses parts of the files that had no changes done to them.  So unless you're expanding out the collapsed sections, they won't be found by "control f" search on the page.  If you take a look at the raw view of the file, or expand out all of the collapsed sections, I *think* you'll find all of the information in question.

Here's a view of the entire raw file submitted in the PR: https://github.com/dmchodnicki/ACVP/blob/e2cca4640229ae7ef27f6b02cf8862ed2952f7e9/src/acvp_protocol.xml

Hopefully that helps!

bfuss:
Got it now. Thanks for the clarification.

PR #45, opened 2017-01-26
Correct Table 1 Formatting

dmchodnicki:
Correct Table 1 formatting in the ACVP Protocol spec.  There were missing cell spacers in the last xml file update.


atvassilev:
Dave,

Per development agreement, when submitting changes to src files you should provide the corresponding artifacts, i.e. the html and txt versions. Could you please fix and resubmit?

dmchodnicki:
Apostol,

     The corresponding changes to the txt and html artifacts were submitted as Pull Requests #46  and #47 respectively.  I didn't see a way to easily include all 3 file updates in one Pull Request.

Issue #41, opened 2017-01-09
Handling of Monte-Carlo tests for symmetric ciphers


atvassilev:
The MC Test for the different modes of symmetric ciphers relies on an algorithm with an initial state (e.g., key, IV, plaintext) - see for example http://csrc.nist.gov/groups/STM/cavp/documents/aes/AESAVS.pdf. The algorithm is different/specific for each mode.

Traditionally, CAVS supplied the initial value for the algorithm and the tested implementation was expected to implement the MCT algorithm, feed the initial condition in it and report the results from the execution. 

The definition of the MCT is not well suited for expanding it into a series of simple test cases because the values for each test depend on the results from the previous test. 

The current specification in sub_symmetric does not appear to have a handle for this type of test. A new type of test is needed for this

@jfigus if you see this differently, please chime in.

atvassilev:
Just to make sure I understand you, are you saying we define an optional type for each tcId? If the type is set and equal to MCT, then this tcId is a Monte Carlo test and the values there are the initial conditions for the corresponding to the mode  MCT algorithm. If the type is not set, the corresponding test case is simple. For example,

{
                               "tcId": 2199,
                               "type" : "MCT",
                               "key": "9dc2c84a37850c11699818605f47958c",
                               "iv": "256953b2feab2a04ae0180d8335bbed6",
                               "pt": "2e586692e647f5028ec6fa47a55a2aab"
},

Similarly, in reporting the results for this tcId, if the type is set to MCT, it contains a list of 100 results from the test, provided in an optional "resultsArray" or something like it. 

"testResults": [
                           {
                               "tcId": 2199,
                                "type" : "MCT",  
                                "resultsArray" : [
                                         {"ct": "1b1ebd1fc45ec43037fd4844241a437f"},
                                         {"ct": "bf43583a665fa45fdee831243a16ea8f"},
                                        .etc.
                                ]
                           },
....
]
In this example I omitted the corresponding key and iv value for each test result but they can be added easily if needed. 

What do you think?

jfigus:
This looks good to me.  Harold may have an opinion on this from the server implementation perspective.

Kritner:
Hey guys, in our discussion yesterday, I believe we were leaning more toward implementing the `testType` within the `testGroup`, rather than the `testCase`.  In addition to the "MCT" type test, there is at a minimum a "KAT", and "MMT" test type. 

[...]

PR #48, opened 2017-01-31
add testType and Informative references

bfussell:
add testType and Informative references
Updated the symmetric cipher list as well.


Kritner:
@bfussell I noticed that the `testType` was added at the vector level in this PR - this differs from the comments in issue https://github.com/usnistgov/ACVP/issues/41#issuecomment-274586401 - was this intended?

PR #55, opened 2017-02-14
Update inEmpty capabilties describption

bfussell:


celic:
If it's alright, I'd like to hold off on this. I'm going to submit a PR making changes to this document and it would be appropriate to handle the change there.

PR #56, opened 2017-02-15
Protocol date fix

celic:
Right now, acvp_protocol.xml only has the year specified for the date tag. For RFCs in the current year, this is automatically converted to a proper date format by XML2RFC. Now that 2016 is over, XML2RFC throws an error and does not parse the file. This adds a month to the date tag to allow XML2RFC to parse the file properly (and uploads the resulting files.)

### DIFF
--- a/artifacts/acvp_protocol.html
+++ b/artifacts/acvp_protocol.html
@@ -850,7 +850,12 @@ Accept:
 "algorithms" : [
          {
 "algorithm" : "AES-GCM",
-"direction" : "encrypt,decrypt",
+"prereqVals" : [{"algorithm" : "AES", "valValue" : "123456"},
+                {"algorithm" : "DRBG", "valValue" : "123456"}],
+"direction": [
+    "encrypt",
+    "decrypt"
+],
 "ivGen" : "internal",
 "ivGenMode" : "8.2.1",
 "keyLen" : [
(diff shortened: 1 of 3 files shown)


bfussell:
I've had issues determining the exact changes in these diffs, so maybe that is the case here.
The diff I see in here shows you've changed the registration/capabilities exchange example from this:

   "direction": [
       "encrypt",
       "decrypt"
   ],

to this:

"direction" : "encrypt,decrypt",

That would not match the acvp_sub_symmetric spec since it is defined as an array.

celic:
I'm not sure how those lines got changed... I was setting up oXygen while making these changes so that might be why more than what is intended was changed... The changes in these files are supposed to only be adding the "month" attribute to the date so that these files compile with XML2RFC. I will fix the other changes to revert them back to master.

celic:
Ok. Changes are done here... In short here is a list of things that needed to be changed

- Date on document needed to be month and year if the year was 2016 for xml2rfc to parse the xml document.
- Expiration date is listed as 6 months after the creation date, so changing the date to "July 2016" made the expiration date early January 2017.
- When just year is listed the expiration date assumes 6 months after the current day (day the file was compiled) or the final day of the year.

So I just edited the date to be 2017 to allow xml2rfc to compile the document. I'm not sure if that sort of date check was in the older version of xml2rfc because according to the metadata there, I am using a newer version. I could not compile the old acvp_protocol.xml file with the new xml2rfc.

Sorry for the numerous changes here trying to fix something simple.

PR #58, opened 2017-02-15
Hash/HMAC Spec Separation

celic:


celic:
Incomplete. Need to finish SHA3/SHAKE documentation.

atvassilev:
I noticed that count is used in the definition of resultsArray for MCT of SHA-3 to enumerate explicitly the array entries. As discussed in the past, using an explicit element counter is unnecessary/redundant and may lead to errors. Please remove it and check for other places where counter may exist.

bfussell:
Were the typos in the Supported Hash Algorithm of the SHA doc and hashmac doc addresses yet ?
In addition I noticed that the testType field for SHA got reverted to the previous version.
We had discussed that the testType field would be KAT or MCT only, no need to differentiate
between short and long, the tests performed are identical.

celic:
There are no KATs for SHA. The only test types are Short, Long and MCT. As you mentioned Short and Long are tested the same way so the field may be redundant there. The only difference is in the generation of the tests. 

I am still working on addressing these changes and should have something up by the end of the week. Normally I'd delete the pull request to avoid someone merging something incomplete, but the comments here are important to remember.

atvassilev:
According to the SHAVS doc at http://csrc.nist.gov/groups/STM/cavp/documents/shs/SHAVS.pdf, the Short Message tests and the Long Message tests are simply KAT's with judiciously chosen test data, i.e. messages of appropriate lengths. 

Wouldn't it make sense to simply call these tests KAT's, similar to what we did for symmetric ciphers? 

To recap the experience there, historically, CAVS had developed several "types" of tests, which were nothing more than just another name for a KAT. On top of this,  this classification had some use for the developers of the CAVS tool but little-to-no-use for the developers of the tested modules. 

It seems to me we are looking at something similar here: from a module developer perspective, I do not see much value in knowing explicitly if a particular test is for a short or long message when computing the digest. Similarly, when verifying the results on the server. This information can be deduced easily from the testId and the value of the specific test message, if needed.   

Again, let's be careful when we look at the CAVS terminology and only port essential information to the new specification.

celic:
I pushed my changes. Please check out the branch if you have a moment to take a look. I addressed Apostol's request on the 'count' field for the MCT array results. Also, for the short and long message tests, I have changed the testType field to accept "functional" as a test type as per #63 even though the issue is not resolved. This allows us to ctrl+f for "functional" and replace it with whatever the new term is.

atvassilev:
Chris, I noticed that in the SHA document you define testType as functional or MonteCarlo, but the examples in Appendix B still have "testType" : "short" and "testType" : "long".. You may want to reconcile this. Also, I like Barry suggestion about using AFT for functional.

celic:
I will make these changes. Also "Algorithm Functional Test" needs to be added to the document instead of just "functional test".

bfussell:
SHA was moved out, so is this section still needed in the hashmac doc ?
2.3.  Supported Hash Algorithms

I still typos in the SHA list in a couple of places, for example:

SHA-512t224
              ^ < --  this "t" ?

celic:
I left Section 2.3 in the HMAC document as it would be important to immediately know which hash functions can be used to validate HMAC implementations.

As for SHA-512t224, the t is for 'truncated' as this is the SHA-512 algorithm with the digest truncated to 224 bits. Within the ACVP code, this mode needed to be labeled as a variable at some point and using "SHA512/224" is invalid. Of course as a string in JSON, "SHA-512/224" is perfectly valid (and consistent with existing documentation.) I can change the document back to this convention.

bfussell:
Section 2.3 just seems redundant if we update 2.4 with the two missing flavors of SHA, 512/224 and 512/256.  Left as is 2.3 would be a bit misleading since it has SHA3 and SHAKE in there and those are not supported by HMAC.

celic:
It seems that SHA3 is supported by HMAC, (via http://csrc.nist.gov/groups/STM/cavp/documents/mac/HMACVS.pdf) but you are right in that SHAKE is not. I will modify the documents to reflect this. I also agree with the redundancy in Section 2.3 and will smooth it out.

bfussell:
You want me to merge what you have and then you do the updates after ?
Doesn't matter to me.

celic:
Might as well group them together. It will be ready tomorrow.

celic:
I made the changes. As I was going through the hash algorithms for HMAC, I added in SHA3 and truncated SHA2. According to the [HMACVS](http://csrc.nist.gov/groups/STM/cavp/documents/mac/HMACVS.pdf) doc, these changes should round out the supported hash functions. 

This change also removes the SHA specific examples from the HMAC document (as they can be found in their separate document already.)

PR #60, opened 2017-02-22
Keying options for tdes

### LINKED ISSUE #59: "KeyingOption" Symmetric Algorithm Capabilities - applies to TDES
It appears the spec would benefit from a "KeyingOption" (or similar) capability for specifying the type of tests to generate for a TDES registration.

The Keying Options as per http://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-20.pdf are described as:

> Three different keying options are allowed by the TDEA. The first option specifies that all the keys
are independent, i.e., KEY1, KEY2, and KEY3 are independent. This is referred to as Keying
Option 1 in FIPS PUB 46-3 – 1999 (and ANSI X9.52 – 1998). It will be referred to as 3-key
TDES in this document. The second option specifies that KEY1 and KEY2 are independent and
KEY3 is equal to KEY1, i.e., KEY1 and KEY2 are independent, KEY3 = KEY1. This is referred
to as Keying Option 2 in FIPS PUB 46-3 – 1999 (and ANSI X9.52 – 1998) and will be referred to
as 2-key TDES in this document. And the third option specifies that KEY1, KEY2 and KEY3 are
equal, i.e., KEY1=KEY2=KEY3. This is referred to as Keying Option 3 in FIPS PUB 46-3 – 1999
(and ANSI X9.52 – 1998 and will be referred to as 1-key TDES in this document. 1-key TDES is
equivalent to single DES.

Keying Options 1 and 2 are the only valid keying options for registration.  Keying Option 3 is still used for performing KATs, but is not specified during registration.

Kritner:
fixes #59

### DIFF
--- a/artifacts/acvp_sub_symmetric.html
+++ b/artifacts/acvp_sub_symmetric.html
@@ -729,6 +729,20 @@
       <td class="left"/>
     </tr>
     <tr>
+      <td class="left">keyingOption</td>
+      <td class="left">The Keying Option used in TDES.  Keying option 1 (1) is 3 distinct keys (K1, K2, K3).  Keying Option 2 (2) is 2 distinct only suitable for decrypt (K1, K2, K1).  Keying option 3 (No longer valid for testing, save KATs) is a single key, now deprecated (K1, K1, K1).</td>
+      <td class="left">array</td>
+      <td class="left">1, 2</td>
+      <td class="left">Yes</td>
+    </tr>
+    <tr>
+      <td class="left"/>
+      <td class="left"/>
+      <td class="left"/>
+      <td class="left"/>
+      <td class="left"/>
+    </tr>
+    <tr>
       <td class="left">ctrSource</td>
       <td class="left">The source of the counter for counter-mode</td>
       <td class="left">value</td>
--- a/artifacts/acvp_sub_symmetric.txt
+++ b/artifacts/acvp_sub_symmetric.txt
@@ -175,49 +175,49 @@ Internet-Draft                Sym Alg JSON                     June 2016
    Each algorithm capability advertised is a self-contained JSON object
    using the following values.
 
-   +------------+--------------+--------------+-------------+----------+
-   | JSON Value | Description  | JSON type    | Valid       | Optional |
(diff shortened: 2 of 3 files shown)


jfigus:
Why include the 1-key option, which is no longer allowed?  Even the 2-key option is only allowed in legacy mode under SP800-131ar1.  Since ACVP is about the future, wouldn't 3-key DES be sufficient?

Kritner:
The Key Option 3 (1 key 3x) while not available for capability registration, would still be used in a KAT scenario for testing.  Does it make more sense to exclude the option in the spec?  When the KAT test vectors are sent back it would look something like:

Key Option 3 (KATS)
```
key1: "K1",
key2: "K1",
key3: "K1",
pt: "..."
```

in Key Option 2 (Decrypt only):
```
key1: "K1",
key2: "K2",
key3: "K1",
ct: "..."
```

Key Option 1:
```
key1: "K1",
key2: "K2",
key3: "K3",
pt: "..."
```

I'm definitely open to making changes, just wanted to get something out there for discussion

hbooth:
@jfigus While we are certainly looking to the future we would like to be sure this service will be able to do everything that CAVs is able to accomplish today.

jfigus:
Is keying option 3 even allowed by FIPS?  If not, why include support for it in the protocol?  Also, do we want to better align the keying option numbers?  It's confusing that keying option 3 uses one key, while keying option 1 uses 3 keys.

bfussell:
I don't believe key option 3 exists in CAVS, it's certainly not on any algorithm sheets I've seen.
Regarding key option 2, CAVS states that beginning 2016 it is non-compliant and will not be tested.

Has something changed there ?

DBugger32:
Keying Option 2 is still valid for decryption.

bfussell:
I stand corrected.  So that is one option we need to add.  

The only question I have is whether option 1 is implied when registering for TDES.
Are there modules that only support option 2 and NOT option 1 ?

DBugger32:
Not in a long time.

Issue #65, opened 2017-03-08
Questions on DRBG 


ashman-p:
Just a few questions on DRBG Test vector handling ...
Take the following example from the spec ...

1. persoStringLen and additionalInputLen are 0 however the fields persoString and additionalInput are populated. Should they be ignored?

2. entropyInputPR and entropyInput will always be the same size?
3. Should "additionalInput"  and "entropyInputPR" be displayed in an array called "predResistanceInput" ?

                {
                  "version": "0.2",
                  "vectorSetId": "1146",
                  "algorithm": "hmacDRBG",
                  "mode": "AES-256",
                  "predResistance": "yes",
                  "entropyInputLen":"256",
                  "nonceLen":"128",
                  "persoStringLen":"0",   ******
                  "additionalInputLen":"0",  **** 
                  "returnedBitsLen":"1024",
                  "testGroups": [
                    {
                      "type": "long",
                      "tests": [
                        {
                          "tcId": "2111",
                          "entropyInput":"ee3392c5f3de6f3f8c4f28d852afacd2cbaa89ed48d1c5d4311662962aa70a98",
                          "nonce":"b991a820fac75fd02642ad8fa651eda4",
                          "persoString":"30f3a50b0e2309dab93ea2aa095e5df8e4b2a42690572b31e53fb79a195481e5",
                          "additionalInput":"4ea46abe95b2e4184113f908ae30123207d481908b7af7ef348657bbf3b4a002",
                          "entropyInputPR": "e4413a2e404f12c644b0b1d7a49a0fbf3d8703d571ffd02168c11b4ade6fc903",
                          "additionalInput" : "61b7204c8fef294f2a9f2e73a83a8a7a04c38e3b3eaaed1b920abfceab912492",
                          "entropyInputPR" : "968ea185d1439fa2d67eb55ac93ba596b1ea679de7c6e44f80dc6f213455f1ed"
                          },
                        {
                          "tcId": "2112",
                          "entropyInput" : "a0ace75784b97224de2957e5f60dc85b25331fcf7901f37418d3c9de17ed4261",
                          "nonce": "b671308068fc7909a360c772f62a4c5e",
                          "persoString" : "338d5f2bd93262da154385e9ed90b7862e3c892f13e1d7d19924b2eb8b3bab21",
                          "additionalInput" : "7acd8bfae17ff4edbac3437817d6b3fce12a04c4034ac6bef0b1b88f7dcd7c85",
[...]

PR #72, opened 2017-03-13
Fixed errors related to ranges of supported parameters

atvassilev:
This PR fixes several errors in the spec and makes improvements in how ranges of parameters are handled - see #65 for details. 

The examples are now consistent among
each other, so that if a particular parameter is set to zero, the corresponding data field is not tested. Please take a look and let me know if you have questions or comments.

Note I did not change the spec to allow for testing multiple intermediate values when additional input is used in prediction resistance mode. This enhancement can be introduced later if the server team is open to it.


atvassilev:
Need to add one more clarification about implementations that can work with and without additional input and personalization data. This would make the spec easier to understand and use.

Issue #67, opened 2017-03-09
Suggestion: "Range" object utilized in spec


Kritner:
Some algorithms may benefit from beginning to use a `range` object that allows for the specifying of a `min` and `max` value that is allowed for the parameter.

AES-CCM and SHAKE specifically use a range of values that are entered.  If 0 min and 32 max are entered in the case of AES-CCM, all values between 0 and 32 are tested.  A random sampling of values within the range is also feasible.

AES-GCM could also use this range object, as currently the `aadLen` array accepts up to 5 values, one can be a 0, 2 can be divisible by 128, and 2 are not divisible by 128.  AES-GCM could begin accepting a min and max, and a separate flag for `supportsNonBlockBoundryAadLen` or similar.

This `range` value will potentially lead to a few new parameters, but should make things a bit more explicit, and hopefully easier to understand.

Example of current registration for AES-GCM:

```
{
	"aadLen": [0, 128, 256, 120, 248]
}
```

could become:

```
{
	"aadLen": {
		"min": 0,
		"max": 256
	},
	"supportsNonBlockBoundryAadLen": true
}
```

Thoughts?

celic:
This would be useful for SHAKE as well, which requires a range of valid sizes for digests.

bfussell:
I agree the min/max range would be an improvement.

atvassilev:
I also like this but think we need to specify also an increment to define it well, provided the entries in the range are equally distant.

Kritner:
Just started a PR #71 

@atvassilev I like the idea of the increment, but could you elaborate on it a bit more?  The use cases I'm seeing currently, are always either an increment of 1, or variable increment.

In AES-CCMs case, we could have 0-65536, where we'd have a 1 increment from 0-32, and then a large jump to the 65536 value.

In AES-GCMs case, we can accept values 0-65536, and the values are always mod 8, but can be either mod 128, or not 128.  In this case, we'd likely just want to randomly select values to test with within the range, that are either mod 128 or not mod 128, depending on the options of the capabilities registration.

@celic I did not make updates to SHAKE as it seems like there's an active PR out for that one #58

atvassilev:
[...]

PR #73, opened 2017-03-14
Rolling back specific "range" changes

Kritner:
Rolled back range changes related to AES-GCM and XTS (they still apply for AES-CCM). 

This PR consists of additional changes related to #67.  Sorry about the confusion, #71 was merged in prior to my being able to change it based on additional feedback.

### DIFF
--- a/artifacts/acvp_protocol.txt
+++ b/artifacts/acvp_protocol.txt
@@ -4,8 +4,8 @@
 
 Internet Engineering Task Force                          B. Fussell, Ed.
 Internet-Draft                                             Cisco Systems
-Intended status: Informational                            March 13, 2017
-Expires: September 14, 2017
+Intended status: Informational                            March 14, 2017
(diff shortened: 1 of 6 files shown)


artifacts/acvp_protocol.txt
```diff
@@ -685,27 +685,25 @@ Internet-Draft              Abbreviated Title                 March 2017
...
-                  },
-   "ptLen" : {
-               "min" : 0,
-               "max" : 256
-                  },
-   "aadLen" : {
-               "min" : 128,
-               "max" : 256
-                  }
+   "ivLen" : [
```
Kritner: This is a GCM example, so it continues to use the literals, rather than range

bfussell:
It seems the example JSON for capabilities in the symmetric spec did not get rolled back.
Had me confused because I thought we still needed to generate the min/max fields.

PR #74, opened 2017-03-14
Fix DRBG parameter range errors and examples

atvassilev:
This PR fixes parameter range errors in the spec. Now with the range construct all possible configurations can be specified for tesing.

Also improved the examples and added dicussion on how to handle special cases of implementations that can work with or without given paramters while the parameters are subject to constraints.


atvassilev:
OK, I'll take care of this shortly. Thanks for the feedback.

Issue #66, opened 2017-03-09
Example JSON, Protocol, Product Registration with Capabilities


meganwoods:
Hi,

The following:

```"operatingSystem" : "Linux 3.1"
     {
     "implementationDescription" : 
          "Sample crypto module for demonstrating ACV protocol."
     },
```
 does not appear to be valid JSON, it is missing a comma then a label.

http://jsonlint.com/ reports:

```
Error: Parse error on line 18:
...ystem": "Linux 3.1" {				"implementatio
-----------------------^
Expecting 'EOF', '}', ':', ',', ']', got '{'
````

**Comment by Kritner** [comment]
@meganwoods it appears some of the JSON from your sample is missing, as the parser error references information not contained within "The following" code block.  

If I'm understanding correctly, this is from the `acvp_protocol` artifact, section `11.2. Product Registration with Capabilities`

Here is the full JSON, assuming we're talking about the similar sections:
```

{
"operation" : "register",
"acvVersion" : "0.2",
"certificateRequest" : "yes",
"debugRequest" : "no",
"production" : "no",
"encryptAtRest" : "yes",
"oeInformation" : {
"vendorName" : "Cisco",
"vendorURL" : "www.cisco.com",
"contact" : "John Doe",
"contactEmail" : "johndoe@cisco.com",
"moduleName" : "Cisco ACV Test Module 1.0",
"moduleType" : "Software",
"operationalEnvironment" : {
"moduleVersion" : "1.0",
"processor" : "Intel Woodcrest",
"operatingSystem" : "Linux 3.1"
     {
     "implementationDescription" : 
          "Sample crypto module for demonstrating ACV protocol."
     },
                            },
                   },
"capabilityExchange" : {
"algorithms" : [
         {
"algorithm" : "AES-GCM",
"prereqVals" : [{"algorithm" : "AES", "valValue" : "123456"},
                {"algorithm" : "DRBG", "valValue" : "123456"}],
"direction": [
    "encrypt",
    "decrypt"
],
"ivGen" : "internal",
"ivGenMode" : "8.2.1",
"keyLen" : [
            128,
            256
             ],
"tagLen" : [
            96
               ],
"ivLen" : [
            96
               ],
"ptLen" : [
            0,
            128,
            136
               ],
"aadLen" : [
            128,
            136
               ]
          }
       ]
   }
}
```

Assuming that the `implementationDescription` is in fact a part of `operationalEnvironment`, the correct JSON could be:
```

{
[...]

PR #75, opened 2017-03-14
Fix for issue #66 that will fix the JSON to be valid

hbooth:


bfussell:
May want to update the .txt and .html files as well.

Issue #82, opened 2017-03-21
Investigate DRBG prediction resistance testing


Kritner:
Wanted to keep this as generic as possible (can update as details get fleshed out).

The spec does not currently account for CAVS parameters `EntropyInputReseed` and `AdditionalInputReseed`.  Both of these parameters were utilized for tests with the following options:

Prediction Resistance Not Enabled = true
Reseeding not Implemented = false, 

There is some potential confusion around why reseeding would be tested in a non PR scenario.  While details and potential solutions are considered, this issue is here to track the conversation.

atvassilev:
Addressed with PR #85

PR #84, opened 2017-03-21
Add option for reseed only prediction resistance

atvassilev:
Added schema enahancements and new examples for supporting reseed-only option. This PR addresses #82.

### DIFF
--- a/artifacts/acvp_sub_drbg.html
+++ b/artifacts/acvp_sub_drbg.html
@@ -400,7 +400,7 @@
 
   <meta name="dct.creator" content="Vassilev, A., Ed." />
   <meta name="dct.identifier" content="urn:ietf:id:draft-ietf-acvp-subdrbg-00" />
-  <meta name="dct.issued" scheme="ISO8601" content="2017-3-20" />
+  <meta name="dct.issued" scheme="ISO8601" content="2017-3-21" />
   <meta name="dct.abstract" content="This document defines the JSON schema for testing DRBG implementations from SP 800-90A  with the ACVP specification." />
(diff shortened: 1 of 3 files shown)


atvassilev:
Typically, the client advertises capabilities with "predResistanceEnabled" : "yes"/"no"; "reseedImplemented": "yes"/"no" flags. Then the server decides based on the logic in the second table in Section 3.1 what test data to prepare for the client. If reseeding is to be tested, the client could look for the presence of reseedInput array. 

I know that the spec provides the flag predResistance but it is also somewhat redundant because one could look for the presence of predResistsnceInput array. A DRBG with no reseeding or prediction resistance capabilities would have neither of the arrays. 

I am more inclined to remove the predResistance flag than to add a reSeed flag. What do you all think?

ashman-p:
I think that, keeping the flags have some more benefits for server and client software because they tend to reduce complexity.
The flags give an explicit contract on what is to be tested and what was sent be the server. So my preference would be to keep the flags.


On Mar 23, 2017, at 8:46 AM, Apostol Vassilev <notifications@github.com<mailto:notifications@github.com>> wrote:


Typically, the client advertises capabilities with "predResistanceEnabled" : "yes"/"no"; "reseedImplemented": "yes"/"no" flags. Then the server decides based on the logic in the second table in Section 3.1 what test data to prepare for the client. If reseeding is to be tested, the client could look for the presence of reseedInput array.

I know that the spec provides the flag predResistance but it is also somewhat redundant because one could look for the presence of predResistsnceInput array. A DRBG with no reseeding or prediction resistance capabilities would have neither of the arrays.

I am more inclined to remove the predResistance flag than to add a reSeed flag. What do you all think?

atvassilev:
OK, then, I will pull this PR back and resubmit it later with both flags added.

Issue #93, opened 2017-03-27
0.3 Symmetric sub spec specifies direction at the vector instead of test_group


hbooth:
I took a look at the .3 spec of the symmetric specification thinking it would be similar to the .2 spec except for some naming differences and I noticed a fairly substantive change that caught me off-guard.

The “direction” JSON field is analogous to the .2 JSON field name “mode” and for some reason it was moved from being a child of a “test_group” object to being in the top-level and a peer/sibling of the “test_groups” field at the test vector level.

bfussell:
Been that way since July it seems.  I don't have a problem moving it wherever you prefer.

hbooth:
In the demo we definitely had it contained in the test_group which greatly simplified a number of issues for us. If we can get it back in there that would be our preference.

atvassilev:
Resolved with PR #95.

PR #94, opened 2017-03-27
Moving direction within TestGroups

atvassilev:
Per agreement, moved direction down to TestGroups. Each test group can
have own direction (encrypt/decrypt). See #93 for details.


atvassilev:
Let me pull it back and fix the protocol spec. This way, it will all be in one PR. I will resubmit shortly.

Issue #96, opened 2017-03-27
acvp.nist.gov -> acvts.nist.gov


hbooth:
Just logging an issue to note that we will be switching from acvp.nist.gov to acvts.nist.gov for the domain name. We will likely want to update the examples to reflect this change. NOTE: The application context will remain the same, so that the URL will be:

https://acvts.nist.gov/acvp/validation/acvp/register

atvassilev:
FYI, acvts stands for Automated Cryptographic Validation Testing Service and we plan to have two application contexts: acvp and amvp. The latter is for module testing, when available.

atvassilev:
Resolved with PR #105.

atvassilev:
Sorry, it should have said PR #102.

PR #97, opened 2017-03-28
Updated URI path prefix and context

atvassilev:
See #96 for details.


atvassilev:
Yeah, I see this and it is rather odd. I have no idea what happened - it is all standard processing with xml2rfc to produce the artifacts. There are no merge conflicts for .xml which is the source. Let me see if I can resolve the conflicts, either manually or by recreating everything from scratch. Do you have any suggestions?

atvassilev:
I resolved the conflicts manually using the web editor but still feel something may be fishy with the files. Pulling this back and will submit shortly, hopefully this time will go better.

PR #105, opened 2017-03-29
Updates for preamble array

bfussell:
One question I have is do we still want to keep these version references:

User Agent: acvp-client/0.3


atvassilev:
I am not sure if acvp-client is used at all on the server-end and does not seem to be communicated back to the client in any of the data sets. My guess is that it can be removed. 

I will merge this PR in and let you deal with acvp-client later.

PR #116, opened 2017-04-06
Add new service url to readme

atvassilev:
Advertice new service URL.


celic:
It seems that an artifact file is causing the merge conflict here. We should be able to just wipe it away and regenerate it.

celic:
I fixed the conflict. Please review the file just to make sure everything looks alright before merging.

It was acvp_sub_sha3.txt

Issue #110, opened 2017-04-03
Utilize "length" specified within TestGroup for TestCases


Kritner:
This might not be a good title, but through integration testing with `AES-GCM` it became apparent that different (CAVS) implementations of algorithms represent a `zero length hex string` differently.  In `AES-GCM`, a zero length plaintext is represented as `""`, whereas in `AES-CCM`, the same is represented as `"00"`.  

It would be beneficial to decide on a consistent way to represent this, and @celic may have some additional input as it relates to `SHA`.  Regardless of the decision, both the client and server should take into account the `PropertyLen` from the `TestGroup` when parsing hex strings.  As it currently stands, that does not appear to be the case.

celic:
My input here is that when reading in the `plaintext` (or `message` for SHA variations), the specified `ptLen` cannot be ignored. There are bit oriented algorithms that will specify a `messageLen` that must be accounted for when reading in the `message`. 

For example...
```
{
    "msgLen": 1,
    "msg": 80
}
```

corresponds to the bit `1` even though the hex could refer to different messages. The length distinguishes it. This is necessary when using hex to convey messages that can be odd (non-multiple of 8) bit lengths. 

With this in mind, as long as the lengths are read in if specified, `"00"` and `""` will resolve to the same value.

atvassilev:
I think we reached a consensus resolution on this issue. Can it be closed?

hbooth:
Consensus maybe, but has anything been updated in the specifications to reflect this consensus? I think we need to make sure that we add something in the documentation with a SHALL/MUST requirement.

atvassilev:
Well how about the submitter of the issue create a PR with what we discussed on the last call and we go from there? 

I think if someone opens an issue and some action/discussion takes place about it, it falls upon that person to follow up with the resolution of the issue unless someone else is assigned to it.  Alternatively, I could start assigning issues to people but this implicit and informal agreement seems preferable to me.

hbooth:
I rather like the idea of explicit assignments. The github issue tracking system supports that. Perhaps whenever our discussions take place we be sure to "assign" someone to get the issue to a resolution. It would make it easier to determine who is responsible for what and avoid something getting dropped or worked on twice.

atvassilev:
OK, I will triage the current open issues and assign those that are clear. Also, we need to add an agenda item for the weekly algotest calls for assignments of current open issues.

Issue #109, opened 2017-04-03
SHA3 Test Cases are Little Endian


celic:
Keccak was designed with little-endianness in mind. The algorithm expects little-endian messages and produces little-endian digests. Unfortunately this is not documented in FIPS-202 and SHA3VS. It should be documented in our spec for SHA3.

PR #134, opened 2017-04-13
value len hex parse

Kritner:
#110

### DIFF
--- a/artifacts/acvp_protocol.html
+++ b/artifacts/acvp_protocol.html
@@ -405,6 +405,7 @@
 <link href="#rfc.section.12.1" rel="Chapter" title="12.1 Resending Test Responses"/>
 <link href="#rfc.section.13" rel="Chapter" title="13 Error Codes"/>
 <link href="#rfc.section.14" rel="Chapter" title="14 Other Considerations"/>
+<link href="#rfc.section.14.1" rel="Chapter" title="14.1 valueLen and hex/bit string parsing"/>
 <link href="#rfc.section.15" rel="Chapter" title="15 Acknowledgements"/>
 <link href="#rfc.section.16" rel="Chapter" title="16 IANA Considerations"/>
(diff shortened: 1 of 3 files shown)


bfussell:
I assume that a valueLen of 5 means the least significant 5 bits ? 
valueLen: 5, value: "FA" is 11010

May be good to point out.

celic:
For most algorithms it is big endian. So in your example, the binary should be "11111". In accordance with issue #109 SHA3 is the only algorithm that is little endian where "FA" would correspond to "11010".

atvassilev:
@celic this is a good point and implies that the valueLen property must also be defined for every algorithm. One approach could be to indicate the endianness problem (for some algorithms it is the LSB part for others it is the MSB part) in the protocol spec and warn the reader to check the algorithm spec for endianness definition. Then each algorithm spec, e.g. SHA3, would define which parts of the bit string to take. 

The bottom line would be to include a couple of examples illustrating the different endians and a warning in the text here for the user to check the corresponding algorithm spec for the endianness.

celic:
Yes, this (should) become apparent when you look at the examples. If it doesn't then there should be additional examples. If you see hex such as "80" for length 1, it is clear the leftmost bit is the significant bit. On the other end, "01" for length 1 shows the opposite. It definitely should also be stated in the specification (ideally as you introduce the length or value the lengths refers to.) As far as I'm aware, SHA3 is the only algorithm that uses little endian hex.

celic:
There is no reference to endianness in this update. The example given "FF" is ambiguous towards endianness as well. Can this be fixed before this is merged?

Kritner:
Ah, I had intentionally left it ambiguous as it was in a general section of the specification, and I thought it was at least on our radar elsewhere for getting more specific with endianness.  

Can make changes (probably tomorrow - need to go drop off my car at the dealership before they close)

celic:
Thanks. As it was brought up in this thread I figured it be worth it to mention in the document now.

PR #135, opened 2017-04-13
More cleanup of SHA from hashmac spec

bfussell:


ashman-p:
Section 4 Table 9 maybe incomplete?

bfussell:
The only values are Mac and Result, Mac is defined in 3.2 but Result is not.
Section 4 points back to 3.2, but since there are only two values I'll add them to section 4
and re-submit.

Issue #151, opened 2017-05-08
Clear definition of what constitutes an "empty" or zero-length string


hbooth:
We need to define clearly what the definition of an "empty" or zero-length string is.

Examples in use:

- ""
- 00
- null
- 0

We should agree to a single definition and be consistent in its use.

atvassilev:
This is likely related to #110. Track accordingly.

atvassilev:
See #110 for details.

Kritner:
I've did a commit at https://github.com/Kritner/ACVP/commit/4c8ecc20aa9d494260a1714933e7f1518b075f0f but have not yet put it into a PR, running through our tests for the suite. 

The gist of the update is to take into account the len when parsing a hex value for a bitstring, but "" is used to signify an empty bit string.  Currently the CAVS implementation is inconsistent between "" and "00" to signify empty bitstring, and we would like to make it consistent on ACVP.  Internally, we seem to agree that "" makes sense, but wanted to run that by everyone.

@bfussell SHA2 and SHA3 currently have some tests that would pass "00" as an empty bitstring, if we make the change to the spec, and our code to pass "", will that impact you?  Should this change instead be slated for .4 as to not impact everything that's already tested in .3?

@hbooth @atvassilev for notifications.

atvassilev:
@Kritner we have agreed no code changes in v0.3. After the demo, we will freeze the v0.3 branch and merge all changes with the master. At that time the changes you propose should be made - I support removing '00' and passing "" instead.

bfussell:
I don’t believe it will impact the client because all the cases where you sent “00”
had a length associated with it and we interpreted the length rather than the “00”.
However, at this late date I’d prefer to hold off on any churn even though it appears
minor.

Barry

Kritner:
@atvassilev should the .3 milestone be removed from #151 and #110 then?  The changes related to the issue would be (potentially) impactful to what has already been tested and signed off on.

atvassilev:
@Kritner  OK, I will changed them to v0.4

PR #158, opened 2017-05-13
Addresses #151 #110

Kritner:
Endianness, hex parsing, empty hex strings


Kritner:
Updated as per suggestion in https://github.com/usnistgov/ACVP/pull/158/commits/253beafde8f62f29a1eb63ac6f1ed21ff1904db5

Issue #160, opened 2017-05-31
Allow RSA Test Generation From Server or From Client


celic:
CAVS allows two ways to validate RSA keyGen. Either CAVS comes up with the inputs and sends them to the IUT to generate output and send back to CAVS to validate (this is similar to most other algorithms); or the IUT generates outputs (and records the inputs) and sends them to CAVS to see if the same result is reached (this is like AES GCM with `ivGen: "internal"`.)

The current RSA spec only allows the latter.

celic:
Fixed with #161

PR #161, opened 2017-06-02
Rsa fixes

celic:
Cleans up RSA spec. Focused on KeyGen. Any changes to SigGen/Ver would be moving hash functions out of the test cases and up to the test group. As well, any strings of integers ("123") were changed to just integers (123).


celic:
Addresses #160

Issue #83, opened 2017-03-21
Suggestion/Poll:  Convert "values" with valid response of "yes/no" to boolean


Kritner:
I know we're getting close to our branch date of the spec, so wanted to get this out there just to get an opinion.  We have several parts of the spec that state an expectation of a value with valid responses being `yes` or `no`; what are everyone's thoughts on using the primitive `boolean` instead?

I'm not (necessarily) suggesting this change be made prior to the branch, but was curious on what everyone's thoughts are - if it's wanted at all, and if it should be done before or after the branch.

hbooth:
I would be supportive of converting to the use of boolean, and I would want to be consistent through out. I can handle changing it now or later.

atvassilev:
I am also supportive of converting to boolean. Timing would depend in part on when the client implementation would like to take the hit on conversion.

atvassilev:
Reclassifying as Enhancement, to help prioritize for branching v0.3.

celic:
Along a similar line here, boolean variables should be named positively. So with `optional = true` this means we do NOT need a variable. While not an immediate problem it is not intuitive. I'd suggest changing the `optional` column to `required` so that `required = true` is a positive affirmation.

celic:
Fixed for SHA, RSA, Symmetric with #161, #162 

This should be kept in mind for other algorithms as we look more in depth at them.

celic:
Needs to be updated:
* protocol
* drbg
* hmac
* kas_ecc
* kas_ffc
* sha3
* sha

celic:
@Kritner was this fixed for HMAC/MAC in #173 ?

Kritner:
@celic yes it was, i forgot to mention that.  Thanks!

atvassilev:
Can this issue be closed now?

celic:
No, there are some documents that haven't been looked at yet that need this conversion. Unless a more specific issue is opened, we need some way to keep track of the remaining documents. As I see, the protocol, kas_ecc, and kas_ffc need this fix but aren't critical issues right now.

celic:
All the algorithms that would be included in v0.4 have been fixed regarding this issue. Its milestone can be changed to "Far Horizons".

atvassilev:
As indicated above, many of the fixes are already in place. I will keep this issue open in the "Far horizons" milestone in case other instances of this coding style are discovered.

PR #162, opened 2017-06-02
Yes/No fixes in SHA

celic:
#83 for SHA algorithms. Checked for Symmetric and this does not appear there.


celic:
Thanks for the heads up.

PR #163, opened 2017-06-05
RSA Mode Fixes

celic:
Fixes some possible redundancies in modes for RSA. Vector sets for KeyGen, SigGen and SigVer should be kept (and generated) separately.

### DIFF
--- a/artifacts/acvp_sub_rsa.html
+++ b/artifacts/acvp_sub_rsa.html
@@ -681,7 +681,7 @@
       <td class="left">algorithm</td>
       <td class="left">The RSA algorithm to be validated</td>
       <td class="left">value</td>
-      <td class="left">"RSA"</td>
+      <td class="left">"RSA-KeyGen", "RSA-SigGen", "RSA-SigVer"</td>
       <td class="left">No</td>
     </tr>
     <tr>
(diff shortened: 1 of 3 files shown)


celic:
In terms of how Test Vectors are generated, the processes are entirely different, (at least for KeyGen and SigGen/Ver). Those two modes could be combined, but even then there is a difference in properties. SigVer needs the added capability of handling SHA-1 (and I believe some additional modulo sizes) for legacy purposes. It would be more clear as a result, if these registrations were handled separately.

atvassilev:
How the processes for specific modes are implemented is very different from what the modes represent. I don't see anything that would prevent someone from implementing any mode with the current layout, there is nothing in it that implies the modes depend on one another.  I don't find the new definition more clear, in fact the opposite. Certainly CAVS treats them that way, but this is not the right way to approach them from a specification point of view.

sam1689:
Hi Guys,

My apologies for being late to the party here...  Have just got to the point where we are implementing RSA algorithm support on our internal server side...

I concede that the RSA modes are not algorithms on their on per se, however in terms of handling these modes of RSA - it definitely matters what the mode is for which you are creating vector sets...

Prior to RSA, we have known at the "top level", where Chris proposed moving the modes, what to prepare for... 

Consequently we could prepare for "AES-GCM" or "SHA-224" at that level... not just "AES" or "SHA" ... sometimes it does not make much difference as in SHA, and you could wait until later to find out the specifics like "SHA-224" ... but for RSA I think it does matter.

In our implementation on the server, so far we have known the specifics up front so to speak... and are structured accordingly.

Conceptually I agree with Chris that if I'm going to do RSA-SigVer I'd like to keep that as its own vector set and not mix RSA KeyGen stuff in with it...

Not sure if that's clear but what I'm trying to say is that there are at least two main reasons to me to push these modes up a level to the "algorithms" section instead of having them down in the "algSpecs -> modeSpecs ->  mode" level...

1. It does fit better with the way we have handled previous algorithms such as AES in my opinion.
2. I find that a more macro-level grouping is preferable for the implementation of these modes.

I think it will actually simplify the amount of information necessary to be specified lower down in the json fields if all applies to the same mode - I think maybe Chris was alluding to that in his explanation.

Again I apologize for being late on commenting on this but I'd like to revisit this if possible.  

Regards,
Sam

celic:
Hi Sam,

I have changes in the works that address these concerns in a slightly different manner. I'll commit them to a different branch (reverting some of the changes made here). Hopefully I can get through this quickly for others to look.

Chris

PR #164, opened 2017-06-06
Fixes unchanged artifact files for SHA-yes/no

celic:
It seems the changes to the artifacts were missing. I just recompiled them and uploaded them here.


celic:
The versions in the examples need to be updated to v0.4

Issue #68, opened 2017-03-09
Update symmetric specification to account for GCM-AES-XPN


Kritner:
GCM-AES-XPN is a relatively new addition, and is not currently accounted for in the specification.  The GCM validation document at http://csrc.nist.gov/groups/STM/cavp/documents/mac/gcmvs.pdf has been updated to include testing information for GCM-AES-XPN.

The additional information needed in GCM-AES-XPN registration is a salt - specifically whether or not that salt is generated internally or externally (similar to the current IV for AES-GCM).  Additionally, GCM-AES-XPN only supports a 96 bit IV, and only supports key sizes of 128 and 256 bits.

PR #165, opened 2017-06-06
Addresses #68 - AES-XPN

Kritner:


Kritner:
@bfussell yeah, there's definitely some work still needed on this spec (I just submit a new issue #166, which I imagine will address most concerns).  I'll do another commit to address the sample json for GCM as you indicated.  

The change to the json in the spec shouldn't have an impact on what's already been developed, looks to just be a miss from before.

PR #171, opened 2017-06-30
DRBG updates

Kritner:
- Updated yes/no to boolean
- removed quotes from numeric values in json
- removed "Range" from property names


Kritner:
#83

PR #172, opened 2017-06-30
Small RSA fixes

celic:
Focuses on reducing redundant properties from RSA-KeyGen and simplifying other existing properties to conform to what they actually refer to. Examples updated as well.

### DIFF
--- a/src/acvp_sub_rsa.xml
+++ b/src/acvp_sub_rsa.xml
@@ -226,20 +226,10 @@
           <c>algorithm</c>
 	  <c>The RSA algorithm to be validated</c>
           <c>value</c>
-          <c>"RSA-KeyGen", "RSA-SigGen", "RSA-SigVer"</c>
+          <c>"RSA"</c>
           <c>No</c>
          <c/>
          <c/>
-          <c/>
-          <c/>
-          <c/>
-          <c>infoGeneratedByServer</c>
-          <c>This flag indicates that the server is responsible for generating inputs for Key Generation tests</c>
-          <c>value</c>
-          <c>true or false</c>
-          <c>Yes</c>
-          <c/>
-          <c/>
           <c/>
           <c/>
           <c/>
@@ -462,6 +452,16 @@
           <c/>
           <c/>
           <c/>
+         <c>infoGeneratedByServer</c>
+         <c>This flag indicates that the server is responsible for generating inputs for Key Generation tests. This flag is not relevant to KeyGen mode "B.3.3" Random Probable Primes</c>
+         <c>value</c>
+         <c>true or false</c>
+         <c>Yes</c>
+         <c/>
+         <c/>
+         <c/>
+         <c/>
+         <c/>
           <c>pubExp</c>
 	       <c>Supports fixed or random public key exponent e</c>
           <c>value</c>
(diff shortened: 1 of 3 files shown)


celic:
No worries. I understand the time is awkward for many. To quickly summarize the changes, some of the KeyGen properties are removed. They are values that were neither input nor output to the functions. Just some intermediary values. After some analysis, looking at CAVS, and talking with some people, we decided they aren't necessary towards the validation. Of note, the other changes are updating the results for B.3.3 KATs to `"passed"/"failed"` instead of `"prime"/"composite"` because the tests don't necessarily check for primality, they check if they pass or fail in the generator. Last one was updating `bitlen1, bitlen2, bitlen3, bitlen4` to be an array as `bitlens` because JSON has no problem supporting arrays.

src/acvp_sub_rsa.xml
```diff
@@ -226,13 +226,23 @@
           <c>algorithm</c>
 	  <c>The RSA algorithm to be validated</c>
           <c>value</c>
-          <c>"RSA"</c>
+          <c>"RSA-KeyGen", "RSA-SigGen", "RSA-SigVer"</c>
```
bfussell: Should the other modes from section 2.1 be listed here ?
Is table 2.1 still accurate ?

src/acvp_sub_rsa.xml
```diff
@@ -226,13 +226,23 @@
...
           <c>value</c>
-          <c>"RSA"</c>
+          <c>"RSA-KeyGen", "RSA-SigGen", "RSA-SigVer"</c>
           <c>No</c>
          <c/>
          <c/>
           <c/>
           <c/>
           <c/>
+          <c>infoGeneratedByServer</c>
```
bfussell: Is this at the proper level ? Seems since it is only related to keyGen that it shouldn't be here.

src/acvp_sub_rsa.xml
```diff
@@ -480,49 +451,29 @@
           <ttcol align="left">Valid Values</ttcol>
           <ttcol align="left">Optional</ttcol>
           <c>mode</c>
```
bfussell: Is this reflected in the examples ? I still see "mode": "keyGen"

src/acvp_sub_rsa.xml
```diff
@@ -1440,23 +1127,27 @@
...
-                        {   "modulo" : 4096,
-                            "hashAlg" : ["SHA-1", "SHA-224", "SHA-256", "SHA-512"]}
-                    ]
-              }]
+         {"modeSpecs" : {
+              "mode": "keyGen",
+              "capSpecs" :  {
+                      "pubExp" : "fixed",
+                      "fixedPubExpVal" : "010001",
+                      "randPQ" : 1,
```
elliemdaw: The other examples got updated, but this should be B.3.2 if I'm not mistaken :)

src/acvp_sub_rsa.xml
celic: You are correct. This should be moved to the TestGroup level for KeyGen only.

src/acvp_sub_rsa.xml
celic: Correct. Thank you.

src/acvp_sub_rsa.xml
celic: I believe this was accidentally reverted to a previous version from my local repo. Thanks for catching it.

src/acvp_sub_rsa.xml
celic: The property name should be "randPQ" not "mode".

celic:
Changes have been made, thank you for the comments.

PR #173, opened 2017-07-05
Mac changes

Kritner:
- Renamed HASHMAC -> MAC - seemed more appropriate as CMAC is contained within doc as well
- Updates bytes to bits
- Moves "len" properties to group, rather than test case
- Updated protocol version
- Removed `inEmpty` property, and made as a part of the array registration for msgLen
- Changed minimum value of "macLen" to be non 0
- Updated json w/ changes to lens
- Update json with sample of verify CMAC operation
- Combined "testGroup" information differences between HMAC and CMAC 
 - renamed properties to keep consistent (msgLen to indicate msgLen for both HMAC and CMAC), rather than having msgLen for CMAC and hLen for HMAC.


Kritner:
added #174 as a question/concern when going through the spec.  There may be other changes needed for TDES and HMAC, I was mostly concentrating on CMAC-AES* for this PR

Kritner:
As pointed out in https://github.com/usnistgov/ACVP/issues/83#issuecomment-313221917 A yes/no value was also converted to a boolean related to #83

PR #176, opened 2017-07-10
Mac aes changes

Kritner:
Additional CMAC changes 
- Added direction property (gen/ver) for CMAC to more closely mirror symmetric cipher properties.
- TestType changed from gen/ver to "AFT" for consistency.
- Updated property types `macLen` and `msgLen` to utilize the Domain object described in usnistgov@17bef01

### DIFF
--- a/artifacts/acvp_sub_mac.html
+++ b/artifacts/acvp_sub_mac.html
@@ -1075,7 +1075,20 @@
                 "macLen": [ 160 ]
             }
             </pre>
-<p id="rfc.section.A.p.2">The following is an example JSON object advertising support for CMAC-TDES.</p>
+<p id="rfc.section.A.p.2">The following is an example JSON object advertising support for CMAC-AES-128.</p>
+<pre>
+
+
+            {
+                "algorithm": "CMAC-AES-128",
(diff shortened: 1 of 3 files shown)


artifacts/acvp_sub_mac.html
```diff
@@ -1049,16 +1075,30 @@ <h1 id="rfc.appendix.A">
...
+<p id="rfc.section.A.p.2">The following is an example JSON object advertising support for CMAC-AES-128.</p>
+<pre>
+
+
+            {
+                "algorithm": "CMAC-AES-128",
+                "prereqVals" : [{"algorithm" : "AES", "valValue" : "123456"},
+		                {"algorithm" : "TDES", "valValue" : "123456"}],
+                "direction" : [ "gen", "ver" ],
+		            "macLen" : [ 64],
```
sam1689: I would fix the white space on 1087 & 1088 - might lend itself to being mis-read :-)

atvassilev: It took me a couple of minutes to understand your comment but I see it now, this applies to the xml source too, not just the html artifact.

Kritner: Updated

PR #185, opened 2017-08-03
Move "mode" back into testGroups for Vector Set Response, fix json typos and minor text typos.

sam1689:
Changes fall into 3 categories:

1.  Updates to position of "mode" field in json examples to comply with the specification.
2.  Fix all json examples which had misplaced commas or other minor typos.
3.  Fix minor text mis-wordings or cut-and-paste errors.

Brief explanation of repositioning of the mode field:

- mode field is essential to creating vector sets and has been positioned within modeSpecs, within algSpecs for registrations
- the current RSA sub-spec indicates that any mode may be registered in a modeSpecs object
- multiple modeSpec objects may be registered and any mode may be chosen in any modeSpec object according to the current spec
- positioning the "mode" globally at the top level in vector set responses from the server would dictate only one mode per vector set - which is contrary to the spec as noted above.  There has been no requirement to have the same mode in each modeSpec object - so this positioning goes against the current spec.
- preserving one vector set per mode would require parsing down and through each modeSpec at the point of registration and would require parsing each modeSpec object twice (1st for registration and 2nd for creating vector sets) - which will not scale well.
- mode field is not needed in every test results response because the server already has the vector set id and test case id which should allow the testGroup information to be pulled.

### DIFF
--- a/src/acvp_sub_rsa.xml
+++ b/src/acvp_sub_rsa.xml
@@ -46,7 +46,7 @@
 
     <!-- Another author who claims to be an editor -->
 
-    <author fullname="Apostol Vassilev" initials="A.V." role="editor" surname="Vassilev">
+    <author fullname="Apostol Vassilev" initials="A." role="editor" surname="Vassilev">
       <organization>National Institute of Standards and Technology</organization>
 
       <address>
@@ -70,7 +70,7 @@
       </address>
     </author>
 
-    <date month="May" year="2017"/>
+    <date month="August" year="2017"/>
 
     <!-- If the month and year are both specified and are the current ones, xml2rfc will fill
          in the current day for you. If only the current year is specified, xml2rfc will fill
@@ -144,7 +144,6 @@
 	    <texttable anchor="algs_table" title="Supported RSA Algorithm Modes JSON Values">
           <ttcol align="left">JSON algorithm value</ttcol>
           <ttcol align="left">JSON mode value</ttcol>
-          <c>"RSA"</c>
           <c>"keyGen"</c>
           <c/>
           <c>"sigGen"</c>
@@ -233,6 +232,16 @@
           <c/>
           <c/>
           <c/>
+          <c>mode</c>
+	  <c>The RSA algorithm mode to be validated</c>
+          <c>value</c>
+          <c>"keyGen", "sigGen", "sigVer", "legacySigVer", "componentSigPrimitive", "componentDecPrimitive"</c>
+          <c>No</c>
+         <c/>
+         <c/>
+          <c/>
+          <c/>
+          <c/>
           <c>prereqVals</c>
 	       <c>The prerequisite algorithm validations</c>
           <c>array of prereqAlgVal objects - see <xref target="prereq_algs"/></c>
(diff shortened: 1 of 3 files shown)


sam1689:
Hi Apostol,
    The switch in your initials was some quirk from doing "make" - no
      idea why it happened - I'll try to fix it when we figure out what
      direction to take on this ...
    I can see the concern about large data files for these modes ...
      I don't have any conceptual problem with doing one mode per Vector
      Set to simplify things and keep the data files manageable... 
    
    If we do that we could remove the redundancy and simplify the
      registration json and all the rest... 
    We can move the mode up to the top level and then we do not need
      the modeSpecs / capSpecs levels...
    
    What do you guys think of something like this:
    {
            "algorithm": "RSA",
            "mode": "sigGen",
            "prereqVals": [{"algorithm": "DRBG", "valValue": "same"},
      {"algorithm": "SHA", "valValue": "same"}],
            "algSpecs" : 
            [
               {
                        "sigType" : "ANSX9.31",
                        "capSigType" :
                        [
                            {"modulo" : 2048,
                              "hashAlgs" : ["SHA-224", "SHA-256",
      "SHA-512"]},
                            {"modulo" : 3072,
                              "hashAlgs" : ["SHA-224", "SHA-256",
      "SHA-512"]},
                            {"modulo" : 4096,
                              "hashAlgs" : ["SHA-224", "SHA-256",
      "SHA-512"]}
                        ]
                    },
               {
                        "sigType" : "PKCS1v1.5",
                        "capSigType" :
                        [
                            {"modulo" : 2048,
                             "hashAlgs" : ["SHA-224", "SHA-256",
      "SHA-512"]},
                            {"modulo" : 3072,
                             "hashAlgs" : ["SHA-224", "SHA-256",
      "SHA-512"]},
                            {"modulo" : 4096,
                             "hashAlgs" : ["SHA-224", "SHA-256",
      "SHA-512"]}
                        ]
                    },
              {
                      "sigType" : "PSS",
                      "capSigType" :
                      [
                          { "modulo" : 2048,
                            "hashAlgs" : ["SHA-224", "SHA-256",
      "SHA-512"],
                            "saltLens" : [28, 32, 64]},
                          { "modulo" : 3072,
                            "hashAlgs" : ["SHA-224", "SHA-256",
      "SHA-512"],
                            "saltLens" : [28, 32, 64]},
                          { "modulo" : 4096,
                            "hashAlgs" : ["SHA-224", "SHA-256",
      "SHA-512"],
                            "saltLens" : [28, 32, 64]}
                      ]
                  }
              ]
         }
    
    Similar simplifications can be made for sigVer and keyGen - I can
      rework the spec along these lines if you guys prefer this way...
    
    Otherwise I think it's best to go with your #2 and I will add
      examples to my PR to clarify that multiple modes could be
      registered/included under one vector set.
    Thanks,
    Sam
    
    
    
    On 8/8/17 1:41 PM, Apostol Vassilev
      wrote:
    
    
      @atvassilev commented on this pull request.
      I understand the point about flexibility in mixing test data
        for various modes in a single testGroup but flexibility comes
        with the cost of complexity and size. If you look at the new
        examples, they still contain test data for a single mode and the
        obvious impression the text leaves with the reader is that
        "mode" is a redundant attribute. To justify it, you would need
        to show examples with test data for two or more modes, but this
        will cause the complexity to increase. Size also matters because
        when you combine data for multiple modes, you may see unexpected
        side-effects related to exchanging large data files.
      So, I see two potential paths forward:
      
        
          Leave the spec as is, taking into account that in reality
            these data sets are likely to be quite bulky so splitting
            the testing per mode may be a practical necessity.
        
        
          Adopt your proposal but rewrite the examples to combine
            data for two or more modes. There is no easy way to do it
            and perhaps you could combine two modes in one example,
            followed by two more modes in another example. Note that
            this would require additional modifications to the document
            to account for combined subsections, etc.
        
      
      One other minor thing, my middle initial is T, not V.
      —
        You are receiving this because you authored the thread.
        Reply to this email directly, view
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
      {"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/usnistgov/ACVP","title":"usnistgov/ACVP","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/usnistgov/ACVP"}},"updates":{"snippets":[{"icon":"PERSON","message":"@atvassilev commented on #185"}],"action":{"name":"View Pull Request","url":"https://github.com/usnistgov/ACVP/pull/185#pullrequestreview-55015068"}}}

celic:
I definitely agree on keeping the specification simple. It seems this would be easiest to accomplish if registrations were handled separately for each mode of RSA. The changes in your comment (not reflected in the PR) are what we should be striving for.

sam1689:
Hi Chris,  

Thanks for the discussion.  

I will adjust the PR to reflect the newer structure above that we agreed on to simplify the registrations.  If it ends up being too cumbersome I may close this pull request and create a new one.

sam1689:
Ok I did an update to this pull request but the diff tool is showing almost every line changes and that was certainly not the case...  Is that what it shows for you guys?   Should I abandon this PR and pull another fork and start fresh?

src/acvp_sub_rsa.xml
```diff
@@ -581,26 +579,36 @@
...
           <c/>
           <c/>
           <c/>
           <c/>
           <c/>
           <c>hashAlgs</c>
-          <c>supported hash algorithms for signature verification - see <xref target="SP800-131A"/>, Section 9</c>
+          <c>supported hash algorithms for this sigType and modulo - see <xref target="SP800-131A"/>, Section 9</c>
           <c>array</c>
           <c>any non-empty subset of {"SHA-224", "SHA-256", "SHA-384", "SHA-512", "SHA-512/224", "SHA-512/256"}</c>
```
celic: Double check and make sure SHA-1 doesn't belong on this list.

src/acvp_sub_rsa.xml
```diff
@@ -619,75 +627,79 @@
...
+          <c>value</c>
+           <c>one of {"ANSX9.31", "PKCS1v1.5", "PSS"}</c>
+          <c>No</c>
+          <c/>
+          <c/>
+          <c/>
+          <c/>
+          <c/>
+          <c>capSigType</c>
+          <c>RSA legacy signature verification parameters  - see <xref target="FIPS186-4"/>, Section 5</c>
```
celic: Legacy changes are incorrect, but those aren't the concern now. These can be fixed later.

PR #186, opened 2017-08-04
Mac updates mac length

Kritner:
- Updates minimum MAC length to 32 bits 
- Specifies maximum MAC length per algorithm in new table

### DIFF
--- a/artifacts/acvp_sub_mac.html
+++ b/artifacts/acvp_sub_mac.html
@@ -650,9 +650,9 @@
 </tr>
 <tr>
 <td class="left">macLen</td>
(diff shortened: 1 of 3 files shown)


Kritner:
@elliemdaw I made a few updates to the "note" section.  They were reflecting when the parameter was of type `array` rather than its current `domain`.  

The json examples can still be considered valid, as with the `Domain` object, you can provide either `Ranges` or `literals` within the `array` (or a combination of both).  Validation authorities may use the provided registration as the basis for certification, and the use of a range of values vs literal values may have an impact on the contents of the certification.

elliemdaw:
@Kritner thank you for the clarification!

PR #192, opened 2017-09-21
Ffc registration updates

Kritner:
+ Refactor in order to accommodate registering KAS in all "modes" of operation within a single registration.  
+ Added JSON samples based on the scheme currently being worked on (dhEphem), will need to add additional as additional schemes are brought in.  
+ Descriptions of the two test types utilized in KAS testing.


Kritner:
I'm working on updating the specification for MQV1 scheme, the sheer number of group and required field possibilities is getting quite large, I'm not sure if there's a better way to represent this or not.

The groups and their requirements for `KdfKc` are a Cartesian product of the following fields:

TestType (2)
* AFT
* VAL

KasRole (2)
* Initiator
* Responder

KcRole (2)
* Provider
* Recipient

KcType (2)
* Unilateral
* Bilateral

so 16 group variants for MQV1 KcKdf, not including the slight differences in MAC type (like aes-ccm nonce).

PR #194, opened 2017-09-25
Changes keyLen to keyingOption and resolves #174

### LINKED ISSUE #174: mac spec keyLen
I noticed that for the mac spec, we're using `keyLen` somewhat inconsistently.

`keyLen` for HMAC is good - we're specifying the bits in the key
`keyLen` for CMAC-AES* does not have any json sample, but assuming would be either 128, 192, or 256.
`keyLen` for CMAC-TDES is (seemingly) being used as the equivalent of "number of keys". In the TDES spec, keyingOption was used to indicate the keys used.  keyingOption 1 = 3 distinct keys, keyingOption 2 = 2 distinct keys, keying option 3 = 1 single key.

Should something similar be done in the mac spec?  Perhaps specify a `keyingOption` property that is valid only for TDES CMAC?  Another option would be to use `CMAC-TDES-keyOption3` and `CMAC-TDES-keyOption2` (or similar) to signify different key options within the registration.  Note for CMAC the `keyLen` will always be the same, it's simply the option that may vary.

pavelgutin:


src/acvp_sub_mac.xml
```diff
@@ -748,19 +753,19 @@
 		"testGroups": [{
 				"testType": "AFT",
 				"direction": "gen",
-				"keyLen": 3,
+				"keyingOption": 3,
```
Kritner: IIRC keying option 3 indicates the same 1 key is used in key1, key2, and key3.  Keying option 1 is the keying option that indicates 3 distinct keys.

pavelgutin: You are correct, I will create a new request with the correct example.

Issue #188, opened 2017-08-11
SaltLen for RSA SigGen PSS needs to be clarified


celic:
Right now we have...

```
"capSigType" :
[
  { 
    "modulo" : 2048,
    "hashAlgs" : ["SHA-224", "SHA-256", "SHA-512"],
    "saltLens" : [28, 32, 64]
  },
  { 
    "modulo" : 3072,
    "hashAlgs" : ["SHA-224", "SHA-256", "SHA-512"],
    "saltLens" : [28, 32, 64]
  },
  { 
    "modulo" : 4096,
    "hashAlgs" : ["SHA-224", "SHA-256", "SHA-512"],
    "saltLens" : [28, 32, 64]
  }
]
```

and it is unclear how the HashAlgs and SaltLens are associated with each other. These are parallel arrays so the i-th index of the HashAlg corresponds to the i-th index of the SaltLens as a valid pairing supported by the client. This should be made clearer, perhaps with an additional object in the registration.

celic:
One suggestion is to build an additional object to capture a hash algorithm and a salt length... The problem with this is the obvious verbosity. In addition to this, this makes the registrations for ANSX and PKCS a bit more complicated.

```
"capSigType" :
[
  { 
    "modulo" : 2048,
    "capabilities": [
      {
        "hashAlg": "SHA-224",
        "saltLen": 20
      },
      {
        "hashAlg": "SHA-256",
        "saltLen": 32,
      },
      {
        "hashAlg": "SHA-512",
        "saltLen": 64
      }
    ]
  }
]
```

This could be expanded very easily to handle fixed salt values during registrations as well as ranges of salt lengths for each hash algorithm.

celic:
This seems to be how the data is handled from the Validation side. See http://csrc.nist.gov/groups/STM/cavp/documents/dss/rsanewval.html#2192

```
[RSASSA-PSS]: 
  Sig(Gen): 
    (2048 SHA( 256 SaltLen( 32 ) , 384 SaltLen( 48 ) , 512 SaltLen( 64 ) )) 
    (3072 SHA( 256 SaltLen( 32 ) , 384 SaltLen( 48 ) , 512 SaltLen( 64 ) )) 

  Sig(Ver): 
    (1024 SHA( 1 SaltLen( 20 ) , 256 SaltLen( 32 ) , 384 SaltLen( 48 ) , 512 SaltLen( 62 ) )) 
[...]

PR #198, opened 2017-10-12
Rsa cleanup

celic:
Addresses #188 and makes the RSA spec more consistent.


celic:
Apostol, I changed them to the original values in the last commit.

PR #200, opened 2017-11-06
adding an optional ctLen and ptLen

pavelgutin:
In some cases, encryption and decryption inputs aren't on the byte boundary. The ctLen and ptLet parameters would allow for a way to support this.


pavelgutin:
This would ensure that all inputs are provided in a hex string format. SHA2 is already using ctLen and ptLen. AES-CFB1 can be modified easily enough to do this as well.

bfussell:
I think a better argument over SHA2 is that AES-CFB1(another symmetric algorithm) is already using binary strings for CFB1.  

Are we making a decision to obsolete bit strings for all current and future algorithms ?

hbooth:
There are two (well three) choices here.

1. Use a bit string wherever we need to represent a sequence of bits that do not necessarily align on a byte boundary. This requires that someone:
 a) know they should process the string as a bit string instead of a hex string
 b) have a bit string processing capability

2. Always represent inputs/outputs as a hex string, but indicate how many bits are "valid". The issue is how to indicate how many of the bits in the hex string should be used. Hex strings everywhere means one processing algorithm, but someone must look for the field that indicates valid bits and truncate appropriately wherever necessary.

3. Use bit strings everywhere. (To be clear, I am not advocating for this.)

If they are any other choices I would be interested in what they would be, but each of these options has something less than ideal about them. During some discussions here it was felt that 2 allowed us to be more consistent and have byte strings everywhere, and in those cases where only some of the bits were valid then the format explicitly stated that. In isolation 1 feels natural for those cases where the algorithm is operating on bits, but now someone needs to know when something is a bit string versus a hex string. 3 I came up with while typing this up (mostly for completeness sake) and while it would allow us to both be consistent and naturally fall back to non-byte aligned bit lengths it is not particularly compact or efficient.

At a minimum if we decide to go with 1 then SHA2 may need to be looked at again to try to make it consistent with that decision. AES-CFB1 will need to be updated if we go with 2 in order to make it consistent.

bfussell:
"a) know they should process the string as a bit string instead of a hex string"
I think the specs should point that out regardless of which way we go, so they should know.

"b) have a bit string processing capability"
We had to create that for AES-CFB1 in the client, so it's there :)

"Hex strings everywhere means one processing algorithm"
One processing function but no less logic since the new function will be required to handle both cases.


I'm not religious about this, however let me point out that if we choose hex strings only any modes that support bit boundaries each and every test case will have to carry the ptLen/ctLen on the request  and the response.

Looks like you are also adding changes to the testType field as well, with "Permutation".  Needs to start lower case.  Is that really a different operation than AFT ?

pavelgutin:
> Looks like you are also adding changes to the testType field as well, with "Permutation"

That actually should be AFT, I grabbed the sample from the wrong file. I will update the pull request.

atvassilev:
I am looking to merge this PR but it reports conflicts with the main line. @pavelgutin  when you have time lets walk through the conflicts to make sure we get the right result.

PR #202, opened 2017-12-04
Ecc registration updates

Kritner:
Applying same updates to ECC as were applied to FFC in #192


atvassilev:
This is a large and detailed PR. I tried to follow it in the order presented by @Kritner. It looks good but only testing will tell if there is some noise left.

PR #203, opened 2017-12-05
Ctr mode updates

celic:
Counter mode tests were updated with different tests. This slightly changes the registration for AES/TDES Counter modes.


celic:
Most of them look the same, but I should provide some example for what the
counter tests look like coming from the server. The registration also
matches other stream ciphers. Will update this today.

PR #206, opened 2018-02-07
Drbg spec updates

celic:
Updates the registration to use more compound objects allowing for multiple 'modes' to be submitted at the same time.

### DIFF
--- a/src/acvp_sub_drbg.xml
+++ b/src/acvp_sub_drbg.xml
@@ -374,17 +374,17 @@
           </texttable>
     
 	
-	         <t>Note 1: If an implementation can be used without prediction resistance, the flag predResistanceNotEnabled shall be set. </t>
-	         <t>Note 2: Implementations that either have prediction resistance always ON or always OFF shall have only one of the prediction resistance flags set. </t>
-	         <t>Note 3: Implementations that can be used either way will have both flags set.</t>
-	         <t>Note 4: For ctrDRBG implementations, at least one of the derivation function options shall be set depending on the one actually used or both</t>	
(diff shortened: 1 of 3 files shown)


src/acvp_sub_drbg.xml
```diff
@@ -741,88 +756,294 @@ the bit lengths the supported values shall be specified explicitly. </t>
...
-                                 "max" : 512,
-                                 "step" : 256} ],
-              "additionalInputLen": [ { "min": 0,
-                                 "max" : 512,
-                                 "step" : 256} ],
-                "returnedBitsLen": 1024
-            }
+{
+  "Algorithm": "ctrDRBG",
+  "Mode": "",
```
Kritner: Remove empty property?

src/acvp_sub_drbg.xml
```diff
@@ -741,88 +756,294 @@ the bit lengths the supported values shall be specified explicitly. </t>
...
-                "reseedImplemented" : true,
-                "entropyInputLen": [ 256 ],
-                "nonceLen": [ 128 ],
-                "persoStringLen": [ 256 ],
-               "additionalInputLen": [256 ],
-                "returnedBitsLen": 1024
-            }
+        {
+  "Algorithm": "hashDRBG",
+  "Mode": "",
```
Kritner: Empty property

atvassilev:
One more thing: the examples list "acvVersion" :  "0.3". This should be corrected to 0.4. Please update the date of the document as well. 

I realize this tedious work but for now we don't have a better mechanism to keep track of these details.

bfussell:
Seems the new examples don't follow the expected camel case used in the definitions.
There are also keyword mismatches, such as prereqVals and the example has PrereqValues.

Issue #209, opened 2018-02-13
Consistency for optional vs required parameters


elliemdaw:
If a field is listed as optional in the spec, it would make sense for the registration to be successful without it being included.

Kritner:
Copying from #210, as I think it is directly related to this issue:

> I'm not sure for properties that are used for only certain algorithms, if it makes sense to say "Optional: No (Required for CMAC-TDES)", or "Optional: Yes (Required for CMAC-TDES)". The idea that I'm attempting to convey is, the property is required for CMAC-TDES, but not relevant for other MAC modes like CMAC-AES and HMAC*

> Perhaps some of the confusion could be alleviated by further breaking out specifications, but am curious what makes the most sense to others.

atvassilev:
It seems this issue was fixed with PR #210.

Issue #208, opened 2018-02-13
CMAC KeyLen parsing issue


elliemdaw:
When registering CMAC:

If "keyLen" is left out, we can an error that the field is missing. When included, the following error is given:

"Incorrectly formatted JSON (95:28): maximum number of values exceeed for field: macLen”

Kritner:
Hey @elliemdaw, assuming this is related to the registration that hit the email group:

(snippet)
```
	"capabilityExchange": {
		"algorithms": [{
				"algorithm": "CMAC-AES-128",
				"prereqVals": [{
					"algorithm": "AES",
					"valValue": "same"
				}],
				"direction": [
					"gen",
					"ver"
				],
				"msgLen": [
					1024,
					0,
					2048,
					0,
					0
				],
				"macLen": [
					64,
					256
				],
				"keyLen": [
					128
				]
			},
			{
				"algorithm": "CMAC-TDES",
				"prereqVals": [{
					"algorithm": "AES",
					"valValue": "same"
				}],
				"direction": [
					"gen",
					"ver"
				],
				"msgLen": [
					1024,
					0,
					2048,
					0,
					0
				],
				"macLen": [
					64,
					256
				],
				"keyLen": [
					128
				]
			}
		]
	}
```

A few things:
* The `algorithm` should actually just be “CMAC-AES”, the lengths are specified in the `keyLen` array property. There was a PR #204 that changed a bit on how the CMAC registration for AES was handled.  This change is not yet reflected within the JSON above or within the driver - we have a ticket in on our end to have the driver end updated for CMAC-AES hopefully within the next few days.
* `macLen` for both AES and TDES exceeds the valid maximum for `macLen` for those algos, should be at a max the block size (so 128 for AES, 64 for TDES)
* `keyLen` doesn’t apply to CMAC-TDES, but `keyingOption` does, and is a required field. 
    * Looking at the specification, this field is in the sample JSON, but is not depicted in the description of the fields, I can apply this update in a PR
* `msgLen` provided values look a bit strange, I’m guessing that the 0s are to equate to how CAVS currently takes input for message lengths?  Would you be able to utilize the “Domain Object” for this sort of registration as described at: http://htmlpreview.github.io/?https://github.com/USNistGov/ACVP/blob/master/artifacts/acvp_protocol.html#rfc.section.14.3 ?

[...]

PR #210, opened 2018-02-13
Cmac tdes keying option

Kritner:
Related to #208

I'm not sure for properties that are used for only certain algorithms, if it makes sense to say "Optional: No (Required for CMAC-TDES)", or "Optional: Yes (Required for CMAC-TDES)".  The idea that I'm attempting to convey is, the property is required for CMAC-TDES, but not relevant for other CMAC modes.

Perhaps some of the confusion could be alleviated by further breaking out specifications, but am curious what makes the most sense to others.  Also related: #209


Kritner:
@atvassilev updated, thanks!

atvassilev:
Looks good, thanks!

PR #213, opened 2018-02-20
Rsa property clean up

celic:
This should be the final set of changes. No content changes, just some property clean-up. Some properties had different names across the modes of RSA. This normalizes them.


celic:
@atvassilev this should be fixed now. We may want to go through other specs then and update that property...

celic:
Not quite ready to merge. Some smaller problems in the text have been pointed out to me. I'm working on fixing them. Most of it is cleaning up some of the paragraphs describing the tables which are now out of date.

atvassilev:
OK, I will hold off with the merge until you submit your corrections. Thanks for the heads up.

celic:
Ok. This should be ready now.

Issue #224, opened 2018-02-27
CMAC definition: keysize of TDES


smuellerDD:
In the symmetric cipher definition, TDES is defined to have 168 bit keys. In the CMAC definition, the keysize seemingly is 192 bits. Please use consistent key sizes.

The CMAC-TDES example in appendix A does not contain a keyLen definition and thus would not be consistent with the spec.

Kritner:
@smuellerDD this appears to be a miss in the spec, TDES does not actually specify a `keyLen` in the registration, as there is only one valid `keyLen` for TDES.  192 bits is the expected `keyLen` (though the keys are split into 8 byte chunks) for TDES, however only 168 of those bits are the "actual key" since there is a parity bit per byte.

I'll issue a PR cleaning up some of the ambiguity in the specification.

Kritner:
I believe this spec discprepancy has been addressed in #225.  If not, feel free to update.

Issue #222, opened 2018-02-26
Symmetric ciphers: do not use multiple data types for one keyword


smuellerDD:
The JSON structures are intended to be processed automatically. To ensure simple code, a particular keyword should not have multiple data types.

For example, ptLen is defined as an array, but it can also be a range for CCM. If it is really needed to define a range, can you please define a different keyword? This should keep the code simple and clean without too many conditionals.

Further, I feel that a range can be equally be represented by an array. Thus, may I propose to disregard the definition of a range entirely?

Kritner:
There is some inconsistency in when using `Array`, `Range` and `Domain`, I believe in most cases the specs should specify `Domain` when a series of valid numeric values is expected.

As described at http://htmlpreview.github.io/?https://github.com/USNistGov/ACVP/blob/master/artifacts/acvp_protocol.html#rfc.section.14.3, the domain can be used to specify literal values, and/or a range of values.  The intention is to give flexibility on the client end, allowing them to either register using a handful of literal values, or the entire range their algorithm implementation supports.

This response seems to apply to multiple recently opened issues: #207, #217 , #221, and #223

smuellerDD:
Am Dienstag, 27. Februar 2018, 15:54:03 CET schrieb Russ Hammett:

Hi Russ,

> There is some inconsistency in when using `Array`, `Range` and `Domain`, I
> believe in most cases the specs should specify `Domain` when a series of
> valid numeric values is expected.
> 
> As described at
> http://htmlpreview.github.io/?https://github.com/USNistGov/ACVP/blob/master
> /artifacts/acvp_protocol.html#rfc.section.14.3, the domain can be used to
> specify literal values, and/or a range of values.  The intention is to give
> flexibility on the client end, allowing them to either register using a
> handful of literal values, or the entire range their algorithm
> implementation supports.
> 
> This response seems to apply to multiple recently opened issues: #207, #217
> , #221, and #223

Thanks for the clarification. But in order to close those items, could you 
please indicate which of the issues use the domain specification and which do 
not?


Ciao
Stephan

Kritner:
I believe this spec discprepancy has been addressed in #225.  If not, feel free to update.

PR #225, opened 2018-02-27
Spec updates consistent types

Kritner:
Handles some spec inconsistencies pointed out in #222 and #224


Kritner:
@atvassilev I made updates for everything except 3, the only difference i can see is where the carriage returns are, they're both the same JSON format.  It seems to be only a display issue, not a content issue.  I would guess that some of the specs json samples weren't run through a linter, but this is something that IMO would be further reaching that just DRBG and the symmetric spec, and (I think) needs further definition either way.

PR #243, opened 2018-03-05
Improve the specification of nonceLen 

atvassilev:
Make it very clear, nonceLen can be at least half the max supported
security strength.


atvassilev:
@Kritner this is a good point. I'll try to come up with something in the next iteration of this spec.

Issue #272, opened 2018-03-20
XTS: IV is marked as "i" in test vectors


smuellerDD:
Please replace the "i" keywrod with "iv" as used in other symmetric ciphers.

Example test vector:
```
[ {
  "acvVersion" : "0.4"
}, {
  "vsId" : 3062,
  "algorithm" : "AES-XTS",
  "testGroups" : [ {
    "testType" : "AFT",
    "direction" : "encrypt",
    "keyLen" : 128,
    "tweakMode" : "hex",
    "ptLen" : 128,
    "tests" : [ {
      "tcId" : 1,
      "key" : "18D61B9C889EFF759B1FBFD114ACA769F9CC66FB14B92E814B9A3E28D07DF993",
      "i" : "66AF129F285BE2D9D4A70B80EC6D475E",
      "pt" : "ED6B21422FC3FE02ADE105D948C69836"
    }, {
```

celic:
`i` is correct here. This does not refer to the IV. It refers to the tweak value. The choice in variable was modeled after CAVS. Perhaps that needs to change, as this issue was brought up previously.

smuellerDD:
Am Dienstag, 20. März 2018, 16:41:01 CET schrieb Chris Celi:

Hi Chris,

> `i` is correct here. This does not refer to the IV. It refers to the tweak
> value. The choice in variable was modeled after CAVS. Perhaps that needs to
> change, as this issue was brought up previously.

It is not documented in the spec. And it would be great to change this keyword 
to something that is more understandable like "tweakValue" or similar.

Note, the tweak is usually handled similar to an IV. A lot of APIs that have 
XTS implemented get their tweak via the IV pointer as XTS does not need a 
"real" IV.

Ciao
Stephan

celic:
Makes sense. I'm totally on board for making the name more understandable (considering this is the second time I'm answering this issue...) I'm not in favor of using `"iv"` because it isn't directly an IV for XTS in the same way other symmetric ciphers use the IV. I do like `"tweakValue"` because that is directly relevant to the context of XTS, and people working with the algorithm would be able to recognize that.

smuellerDD:
Am Dienstag, 20. März 2018, 17:22:41 CET schrieb Chris Celi:

Hi Chris,

> Makes sense. I'm totally on board for making the name more understandable
> (considering this is the second time I'm answering this issue...) I'm not
> in favor of using `"iv"` because it isn't directly an IV for XTS in the
[...]

Issue #271, opened 2018-03-20
TestGroup Property testType missing for XTS and other Algs


smuellerDD:
For XTS vectors, the testType is missing

celic:
They are all AFT. This might happen on more algorithms that only have a single test type though...

smuellerDD:
Am Dienstag, 20. März 2018, 16:42:17 CET schrieb Chris Celi:

Hi Chris,

> They are all AFT. This might happen on more algorithms that only have a
> single test type though...

May I request to have that added to ensure a full machine-parsable JSON data 
set is issued? To ensure efficient machine-readability, common structures 
should always be adhered to.

Thanks.

Ciao
Stephan

celic:
I agree. This is on our agenda internally. I'll leave this ticket open for now. This may not be an immediate fix unfortunately. If you see a test group without a testType, assume that it is an AFT.

smuellerDD:
Am Dienstag, 20. März 2018, 17:19:15 CET schrieb Chris Celi:

Hi Chris,

> I agree. This is on our agenda internally. I'll leave this ticket open for
> now. This may not be an immediate fix unfortunately. If you see a test
> group without a testType, assume that it is an AFT.

I have developed very generic parser which is agnostic of any data, but knows 
the information structure. This parser is fed with a data model for the 
different cipher specifications. Thus, lacking such structural differences is 
not trivial to handle in my parser.

Since we are in a test mode, I can handle to manually add the AFT though. But 
I am asking for structural consistency before the release.

Ciao
Stephan

celic:
If you notice it missing in an algorithm, let's build a list under this issue to make sure everything gets covered.

smuellerDD:
Am Dienstag, 20. März 2018, 18:14:34 CET schrieb Chris Celi:

Hi Chris,

> If you notice it missing in an algorithm, let's build a list under this
> issue to make sure everything gets covered.

So far, I did not come across any other example. But I will note it if I spot 
such an issue.

Ciao
Stephan

smuellerDD:
ECDSA seems to have no testType (according to the spec).

celic:
All algos should have testType now with our recent JSON parser changes.

Issue #250, opened 2018-03-07
AES-CTR/TDES-CTR uses different pt/ct string than all other symmetric ciphers


bfussell:
pt/ct changed to plainText/cipherText. all the other symmetric ciphers use the former.
I missed it when I reviewed and approved the PR :)

atvassilev:
Resolved with PR #274.

PR #274, opened 2018-03-20
Update all AES examples, except XPN

bfussell:
Examples  reflect recent issues with plainText/cipherText(#250), tweakValue(#272) and AFT(#271).
No examples for XPN since I can't generate them.


bfussell:
Sure, just wasn't sure if we had closure on that.  Let me join the discussion on it.

PR #296, opened 2018-03-29
Ecdsa changes

celic:
Lots of changes for ECDSA similar to DSA. Don't think there are any errors in this so please review away!


celic:
Sorry about that. I forgot to rename some of the examples I pasted around so there was a compilation error in generating the artifacts. This has been fixed and uploaded.

PR #313, opened 2018-04-04
Updates TLS

celic:


Kritner:
Is anything waiting on this PR to be merged?  Probably need to take into account this modified spec when making changes for consistent casing for SHA (unless its already done, I have not checked)

celic:
Yeah. We want to look at the upcoming version of SP 800-56C to see if we can quickly update this to handle the new TLS KDF, or if that would take more work, merge this request as is.

atvassilev:
We will look into the TLS coverage later.

Issue #338, opened 2018-04-20
DSA pqgGen: server failed to generate vector


smuellerDD:
The server failed to generate pqgVer vectors with the error:

```
[ {
  "acvVersion" : "0.4"
}, {
  "vsId" : 2136,
  "status" : "error during vector generation",
  "message" : "failed to generate tests for: 2136"
} ]
```

My request was:

```
        {
          "algorithm":"DSA",
          "mode":"pqgGen",
          "prereqVals":[
            {
              "algorithm":"SHA",
              "valValue":"same"
            },
            {
              "algorithm":"DRBG",
              "valValue":"same"
            }
          ],
          "capabilities":[
            {
              "l":3072,
              "n":256,
              "PQGen":[
                "probable"
              ],
              "GGen":[
                "unverifiable"
              ],
              "hashAlg":[
                "SHA2-256"
              ]
            }
          ]
        }
```

Kritner:
Same issue as: https://github.com/usnistgov/ACVP/issues/337#issuecomment-383115695

Kritner:
The fix has been deployed, please update this issue if there any problems.  Closing for now.

smuellerDD:
Am Donnerstag, 10. Mai 2018, 19:50:48 CEST schrieb Russ Hammett:

Hi Russ,

> The fix has been deployed, please update this issue if there any problems. 
> Closing for now.

Currently I have heavy authentication issues I need to sort out with Harold. 
Furthermore, just right now, I get a 404 error from the server even when 
trying to log on.

Ciao
Stephan

smuellerDD:
Am Donnerstag, 10. Mai 2018, 19:52:21 CEST schrieb Stephan Mueller:

Hi,

> Am Donnerstag, 10. Mai 2018, 19:50:48 CEST schrieb Russ Hammett:
> 
> Hi Russ,
> 
> > The fix has been deployed, please update this issue if there any problems.
> > Closing for now.
> 
> Currently I have heavy authentication issues I need to sort out with Harold.
> Furthermore, just right now, I get a 404 error from the server even when
> trying to log on.

It seems that is cleared now and was due to the server update. Please 
disregard my earlier message.
> 
> Ciao
> Stephan



Ciao
Stephan

Kritner:
Thanks @smuellerDD will do

smuellerDD:
Sorry, but server still reports:

```
[ {
  "acvVersion" : "0.4"
}, {
  "vsId" : 55,
[...]

Issue #337, opened 2018-04-20
DSA pqgVer: server failed to generate vectors


smuellerDD:
The server failed to generate pqgVer vectors with the following message:

```
[ {
  "acvVersion" : "0.4"
}, {
  "vsId" : 2137,
  "status" : "error during vector generation",
  "message" : "failed to generate tests for: 2137"
} ]
```

My request was:

```
        {
          "algorithm":"DSA",
          "mode":"pqgVer",
          "prereqVals":[
            {
              "algorithm":"SHA",
              "valValue":"same"
            },
            {
              "algorithm":"DRBG",
              "valValue":"same"
            }
          ],
          "capabilities":[
            {
              "l":3072,
              "n":256,
              "PQGen":[
                "probable"
              ],
              "GGen":[
                "unverifiable"
              ],
              "hashAlg":[
                "SHA2-256"
              ]
            }
          ]
        },
```

Kritner:
working on internal fix - this is due to inconsistencies in "hashAlg" vs "hashAlgs".  My current intention is to use "hashAlg" across the DSA board.  This will require a spec update as some sample json currently uses "hashAlgs".

Kritner:
The fix has been deployed, please update this issue if there any problems.  Closing for now.

smuellerDD:
confirmed, working.

Issue #335, opened 2018-04-20
DSA pqgVer: not supported for L=1024/N=160?


smuellerDD:
When I request the pqgVer for L-1024, I get the following response from the server:

```
[ {
  "acvVersion" : "0.4"
}, {
  "vsId" : 2141,
  "status" : "error during vector generation",
  "message" : "failed to generate tests for: 2141"
} ]
```

My request was

```
        {
          "algorithm":"DSA",
          "mode":"pqgVer",
          "prereqVals":[
            {
              "algorithm":"SHA",
              "valValue":"same"
            },
            {
              "algorithm":"DRBG",
              "valValue":"same"
            }
          ],
          "capabilities":[
            {
              "l":1024,
              "n":160,
              "PQGen":[
                "probable"
              ],
              "GGen":[
                "unverifiable"
              ],
              "hashAlg":[
                "SHA-1",
                "SHA2-224",
                "SHA2-256"
              ]
            }
          ]
        },
```

Kritner:
Appears to be same issue as https://github.com/usnistgov/ACVP/issues/337#issuecomment-383115695

Kritner:
The fix has been deployed, please update this issue if there any problems.  Closing for now.

smuellerDD:
Confirmed, working

Issue #334, opened 2018-04-20
DSA sigver: PQG missing?


smuellerDD:
The test vectors for DSA sigVer contain

```
{
      "tcId" : 1,
      "message" : "77250571BC47EA07BE1D1B700C2BFF5526B53B96",
      "y" : "8C1BBD4586A31AC43FB4C16315F509306837E5B8BC62091CC663D5301326F094832175C1BB19E1F35D4438E11B07B714CBD71C0364E5051F0B9F3ABE418971CE3CD667E09243828279A27D5E470F2E8B82431DA31537F3D0D81D12321CDCBFE861A08D71A45DDE741A611ECE686F33F8F3CB4E35BC30DFCEAF0793F49E5751C9",
      "r" : "94B7ED20B5F69CF3AADDD18CA73D160A89B8CD04",
      "s" : "6D19A5CB5490547A903381E6A30D1930CF54F7D8"
    }
```

I wonder whether the PQG values are missing here? Or how shall an IUT come by these values?

Kritner:
@smuellerDD agreed, it seems that P, Q, and G should be included as group level properties.  I will move forward with an internal fix.  @celic are you confirm this fix is valid?  Seems that sig verify requires the PQG as a parameter, but we not currently including that in our projections to the IUT.

Kritner:
The fix has been deployed, please update this issue if there any problems.  Closing for now.

smuellerDD:
Confirmed, working.

PR #340, opened 2018-04-20
Dsa hashAlgs -> hashAlg, and pqg

Kritner:
Spec updates related to #334, #335, #337, #338


celic:
Thanks for handling this @Kritner, this looks ready to merge @atvassilev

artifacts/acvp_sub_dsa.html
```diff
@@ -1673,7 +1709,7 @@ <h1 id="rfc.appendix.A.1">
       ],
       "L": 2048,
       "N": 256,
-      "HashAlgs": [
+      "HashAlg": [
```
bfussell: HashAlg or hashAlg  ? I think we use the latter in most algorithms

artifacts/acvp_sub_dsa.html
```diff
@@ -1691,7 +1727,7 @@ <h1 id="rfc.appendix.A.1">
       ],
       "L": 3072,
```
bfussell: L , N or l, n ? Seems like the implementation is using lower case for both.

celic:
All the examples should be the agreed camelCase.

bfussell:
I was looking at an outdated PR. I deleted my comments, everything looks good.

PR #342, opened 2018-04-27
Modify the case for the SHA algorithm to be lowercase everywhere

pavelgutin:
There were some inconsistencies between the capitalization of SHA, so this makes everything lowercase.


Kritner:
@pavelgutin I would say it would be more consistent to be upper cased everywhere, because now while SHA is consistent, it's now inconsistent with other pre-req algos used like AES and/or TDES.

atvassilev:
I second  @Kritner . Moreover, the SHA algorithms are always referred to in capital letters in the FIPS and Special Pubs documents. For example FIPS 202.

Kritner:
Closing for now, have separate PR #345 that capitalizes sha (among other algorithms for consistency)

PR #345, opened 2018-05-01
Algo name consistency

Kritner:
updates casing to be consistently upper case for "SHA", "HMAC", "CMAC", "TDES", "AES"


Kritner:
@hbooth updates to SHA, HMAC, CMAC, AES, TDES casing for consistency.  Relates to #342

PR #355, opened 2018-05-18
Kas updates component breakout

Kritner:
Breaking `NoKdfNoKc` modes of KAS-ECC and KAS-FFC into their own "algorithm" and "mode".

`KdfNoKc` and `KdfKc` continue as they were, registered under "KAS-FFC" or "KAS-ECC".
`NoKdfNoKc` as a component validation, is now registered separately under:

* Algorithm: "KAS-ECC" or "KAS-FFC"
* Mode: "Component"


src/acvp_sub_kas_ecc.xml
```diff
@@ -1400,7 +1392,7 @@
 				"kdfType": "asn1",
 				"idServerLen": 48,
 				"idServer": "434156536964",
-				"curveName": "p-256",
+				"curve": "p-256",
```
celic: always prefer capital for P, B and K in curve name?

celic: Could probably be extracted to a separate issue than the one this is intending to fix.

Kritner: Perhaps the capitalized curve names could be in a separate PR? ECDSA is currently using lowered curves names as well, so at the least they're currently consistent.

Issue #64, opened 2017-03-03
Provide algorithm/capability query service


atvassilev:
As the list of approved algorithms evolves over time, it would be nice to a have an ACVP server query that would return all approved algorithms/modes available for testing at a particular moment in time.

Under the current model, the client registers capabilities with the server with the underlying assumption that both the client and server know the valid options. However, it is conceivable in some cases the client may need to discover what is available for testing. Such a query would provide the needed information to the client.

Issue #220, opened 2018-02-26
ACVP spec: cpe keyword


smuellerDD:
The ACVP protocol defines the keyword "cpe" without seemingly defining what shall be communicated with this keyword. Can you please elaborate?

Kritner:
This seems to have been covered in #392, closing.

Issue #219, opened 2018-02-26
ACVP spec: "feature" data type


smuellerDD:
The ACVP protocol defines the keyword "features" which seems to be some kind of flags type. If so, which particular values are allowed?

If it is a free-form data field, please add that hint to the documentation.

Kritner:
Resolved from #392 as per PR message

Issue #146, opened 2017-05-04
Include the retry in /results when disposition is incomplete


atvassilev:
Enhance the description in Section 11.5 of the protocol spec to describe the returned retry when the disposition is incomplete.

celic:
@hbooth this is done correct?

PR #392, opened 2018-06-06
Initial Changes for 0.5 Interface

hbooth:
Creating this pull request in order to facilitate discussion around it and allow external folks to comment. This pull request will resolve: #64, #111, #117, #118, #145, #146, #219, #220.

The only file of substance which has changed is src/acvp_protocol.xml but you can easily view the final html artifact through:

http://htmlpreview.github.io/?https://github.com/usnistgov/ACVP/blob/wip_restapi_changes/artifacts/acvp_protocol.html

Comments and discussion can occur on this pull request.


hbooth:
We would like to merge this branch on June 22. Please provide any comments prior to June 22 so we can integrate any changes.

PR #413, opened 2018-07-02
Kas updates

Kritner:
Various updates (see individual commits for further details) on KAS specifications


atvassilev:
I am about to merge this PR. Note that this may require changes to the Web Server JSON parser because a new property kasRole is introduced. A definite code change is required for the gen/val back-end.

Kritner:
Thanks @atvassilev, we're testing a change internally for the `role` -> `kasRole` change (among other things), so once that change actually does make it out to prod, will match the specification.

Until that push actually goes out however, `role` will still be used in the registration.

atvassilev:
@Kritner this is cool! Please close Issue #415 when you guys are done with testing.

PR #415, opened 2018-07-05
workflow description with JWT

atvassilev:
A needed enhancement for the JWT spec and workflows is added.

### DIFF
--- a/artifacts/acvp_protocol.html
+++ b/artifacts/acvp_protocol.html
@@ -461,12 +461,12 @@
 <link href="#rfc.authors" rel="Chapter">
 
 
-  <meta name="generator" content="xml2rfc version 2.9.6 - https://tools.ietf.org/tools/xml2rfc" />
+  <meta name="generator" content="xml2rfc version 2.9.9 - https://tools.ietf.org/tools/xml2rfc" />
   <link rel="schema.dct" href="http://purl.org/dc/terms/" />
 
   <meta name="dct.creator" content="Fussell, B., Ed." />
   <meta name="dct.identifier" content="urn:ietf:id:draft-ietf-acvp-protocol-0.4" />
-  <meta name="dct.issued" scheme="ISO8601" content="2018-06" />
+  <meta name="dct.issued" scheme="ISO8601" content="2018-05" />
   <meta name="dct.abstract" content="This document defines the Automated Cryptographic Validation Protocol(ACVP)." />
   <meta name="description" content="This document defines the Automated Cryptographic Validation Protocol(ACVP)." />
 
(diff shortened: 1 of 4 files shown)


hbooth:
Is the intent to only provide some limited context about the JWT workflow and point to the secondary document or to fully describe what's needed to obtain and use a JWT? If the latter then I think more information will be necessary.

The table is difficult to read, it may be better to, in the column that is the destination for the communication (or maybe a column to the right of the text) to indicate the direction of communication.

hbooth:
https://github.com/usnistgov/ACVP/blob/fe97357026d0c0a791d7a532e55379b7dd7f3311/src/acvp_protocol.xml

atvassilev:
The intent is to provide a limited but useful set of examples that allow the reader to infer how to work with all RESTFul API's available. Given that authentication is out of scope for this spec, the exact specification of authentication flows that return a JWT token is not possible. That is why this step is sketched out only. The workflow description can grow over time if we determine that some critical examples are missing but hopefully, it will not evolve to a full description. If we identify that we need a complete specification, then the content of the services table would have to change. 

Since it we are limited to text only, it is hard to make an easy-to-read table, just look at the pictures in the same spec. The presentation in this table is consistent in style with the graphics in the picture. I added a verb on top of each line to help indicate the direction of communication more prominently,  i.e. 'retrieve' for the returns. In addition, two types of lines are provided, one with wider dash space than the other. Did you overlook some of these elements in your reading of the table?

I am not convinced that adding an extra column would help better.

atvassilev:
BTW, I just looked at this again and realized that the text and html files look differently. Are your comments based on viewing the html or txt?

hbooth:
I was looking at the HTML version.

atvassilev:
Take a look at the txt version, you can view it directly from the browser: https://github.com/usnistgov/ACVP/blob/inlcude-authorization-flows-in-protocol-spec/artifacts/acvp_protocol.txt 

Let me know what you think of that presentation.

hbooth:
The .txt version is easier to follow than the html, but perhaps we could still consider ways to improve the table so that .html version and .txt can both be interpreted with the same ease.

atvassilev:
Apparently, it is not easy to reconcile html and txt. For example, the services table in txt does not look very good, in html it looks fine. 

I am going to try to convert my table into a figure - seems to hold up better the conversion to various formats. The services table is a different matter, you may have to find another way to make it look better in txt.

atvassilev:
@hbooth I just committed changes that convert the text table into a figure. This shows up ok in html and txt on my machine. Boy, what a pain it is to work with RFC formatting! Can you confirm it displays fine in html for you?

atvassilev:
I just pushed a couple of additional refinement that somehow got left out when I converted the text table to a figure to avoid messing up the presentation in the two formats: txt and html. A clarification that all authorization flows are over HTTP. Bumped the spec version to 0.5. 

I cannot help but vent one more time about the primitiveness of the tool I used to convert my file to txt and html. Looking forward to the time we work on the spec for IETF, it may be useful to look for other alternatives, if available.

atvassilev:
I reverted one unnecessary commit (temporary auto commit on the drbg spec) courtesy of my old GitHub client. This was raising conflicts for merging this PR into the master. It is clean now and contains only the protocol-related changes.

BTW, if you are using a desktop Git client, not just a command shell, the latest available version is much better than the old one.

Issue #395, opened 2018-06-11
CTR DRBG: test vectors without PR seem wrong


smuellerDD:
I get the following vectors for predictionresistance == false:

```
    "testType" : "AFT",
    "mode" : "AES-128",
    "derFunc" : true,
    "predResistance" : false,
    "entropyInputLen" : 128,
    "reSeed" : true,
    "nonceLen" : 64,
    "persoStringLen" : 0,
    "additionalInputLen" : 0,
    "returnedBitsLen" : 512,
    "tests" : [ {
      "tcId" : 1,
      "entropyInput" : "CAED3A22E6CEF0801CB84C393CAD1B10",
      "nonce" : "214AC39FACDA5E84",
      "persoString" : "",
      "otherInput" : [ {
        "additionalInput" : "",
        "entropyInput" : "5DBF858FCF24267F3C1CBFD2B3CAF805"
      }, {
        "additionalInput" : "",
        "entropyInput" : ""
      }, {
        "additionalInput" : "",
        "entropyInput" : ""
      } ]
    }, {
```
The providing of otherInput seems wrong as we do not have prediction resistance. Besides, there are 3 entries in otherInput.

celic:
Just to confirm, the otherInput is not used when prediction resistance is not enabled?

smuellerDD:
Am Donnerstag, 21. Juni 2018, 20:21:38 CEST schrieb Chris Celi:

Hi Chris,

> Just to confirm, the otherInput is not used when prediction resistance is
> not enabled?

Here is the DRBG test definition:

 For cases with prediction resistance enabled, each trial consists
of (1) instantiate drbg, (2) generate ReturnedBitsLen random bits, do not 
print (3) generate
ReturnedBitsLen random bits, print out (4) uninstantiate.

==> Before each generate, the entropy / additional info must be added.

For test cases without prediction resistance, each trial consists of (1) 
instantiate drbg, (2)
generate ReturnedBitsLen random bits, do not print (3) reseed, (4) generate
ReturnedBitsLen random bits, print out (5) uninstantiate.

==> the reseed implies that entropy / additional info must be added.

The test vector now contains 3 entries in otherInfo. So, I would rather think 
that for PR=false and reseed=true, otherInput should have 1 entry (i.e. the 
reseed entropy/addtl info).

Please see the DRBG CAVS spec section 6.2 for the different options and input 
data.

Ciao
Stephan

celic:
[...]

PR #416, opened 2018-07-05
Indicate intended use explicitly

atvassilev:
This update provides detailed discussion of various test options for DRBG implementations with various reSeed and predResistance options. 

It also introduces an explicit property "intendedUse" in otherInput objects to indicate to the client how to use the test data contained in the additionalInput and entropyInput properties. 

This addresses issue #395.

### DIFF
--- a/src/acvp_sub_drbg.xml
+++ b/src/acvp_sub_drbg.xml
@@ -624,18 +624,11 @@ the bit lengths the supported values shall be specified explicitly. </t>
 		<c>array of additonal input/entropy input value pairs for testing. See <xref target="vs_prtc_table"/></c>
 		<c>array</c>
 		<c>No</c>
-    <c/>
-		<c/>
-		<c/>
-		<c/>
-		<c>intendedUse</c>
-		<c>"both", "reSeed", "predResistance", "other"</c>
-		<c>value</c>
-		<c>No</c>
 	    </texttable>
 
-	    <t>Each prediction resistance test group contains an array of one or more test vectors, typically two.  Such tests are
-	     generated when the implementation under test supports prediction resistance. Note that in this case both tests have non-empty entropy input.  
+	    <t>Each test group contains an array of one or more tests.  Each test object contains an otherInput object,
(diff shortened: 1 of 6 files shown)


Kritner:
I'll take a look at this today.  @atvassilev it looks like changes from #415 made it into this PR as well

atvassilev:
@Kritner, this must be some kind of a mix up with my local Git setup. Try to ignore these changes if you can. I'll look to see what went wrong.

src/acvp_sub_drbg.xml
```diff
@@ -624,9 +624,19 @@ the bit lengths the supported values shall be specified explicitly. </t>
 		<c>array of additonal input/entropy input value pairs for testing. See <xref target="vs_prtc_table"/></c>
 		<c>array</c>
 		<c>No</c>
+    <c/>
```
Kritner: I think this would make more sense as a property on the `otherInput` object itself rather than the test case?  The `otherInput` entry will always be either applied to a `reseed` or `generation` operation.

So it could look like this...

```
otherInput: [
     {
          "intendedUse" : "reseed",
          "additionalInput" : "CAFECAFE",
          "entropyInput" : "CAFECAFE"
     },
     {
          "intendedUse" : "generate",
          "additionalInput" : "CAFECAFE",
          "entropyInput" : "CAFECAFE"
     },
     {
          "intendedUse" : "generate",
          "additionalInput" : "CAFECAFE",
          "entropyInput" : "CAFECAFE"
     },
]
```

Kritner: oh and additionally `reseed` will always be the first operation when there are 3 entries.  If there are only 2 entries, those two entries are always `generate`

atvassilev: @Kritner I guess I was trying to save some precious bytes to not overburden the communication between the client and the server - recently we have been hitting the limits. However, I see your point and like it. I will try to update my PR - keep your fingers crossed it will work properly this time.

atvassilev:
@Kritner I just pushed an update to this PR along the lines above. Take a look and let me know what you think.

PR #420, opened 2018-07-12
JWT expiration/renewal

atvassilev:
Added a section on JWT expiration/renewal. Added the people who have contributed to the spec as authors.


atvassilev:
@celic, yes, this is an interaction diagram and typically they are left open at the bottom. I guess a matter of technical etiquette...

PR #427, opened 2018-08-06
An initial framework for algorithm test extensions

atvassilev:
The minimum required set to explain this consists of three specs: the protocol,
the new algorithm test extension spec, the symmetric block cipher spec.

This is only the first iteration on this and as such it is incomplete in
elaborating fully on all the necessary text. However, it allows the effort
to be partitioned between the editors for the three documents.

Note also that the new specs are renamed to match IETF guidelines for
Internet drafts.


artifacts/draft-fussell-acvp-spec-00.txt
```diff
@@ -0,0 +1,2912 @@
...
+                 |                         |                         | |
+                 |                         |      DUT                | |
+                 |                         +-------------------------+ |
+                 | Vendor/Customer Premise                             |
+                 +-----------------------------------------------------+
+
+
+                                 Figure 2
+
+4.  ACV Protocsubol
```
bfussell: typo, ACVP Protocol

artifacts/draft-fussell-acvp-spec-00.txt
```diff
@@ -0,0 +1,2912 @@
...
+   protocol, e.g., HTTP, custom server, context and port number to
+   interact with the DUT.
+
+4.2.  HTTP URI Resources
+
+   In the table below, any part of a resource path enclosed in curly
+   braces, { or }, are replaced by an instance of what is described in
+   the brackets.  For example {testSessionId} could be replaced with 1.
+
+     _An empty cell for a resource and HTTP Method combination denotes
```
bfussell: is there something in the xml that leaves some underscores in the text version ? Here and below (405)"._

atvassilev: @bfussell thanks for the detailed review. Several of the errors you point out are leftovers from the prior text that I did not change because my focus was on the new content. They are valid but will be for the editors of each new spec to address. I will take care of your feedback on draft-vassilev-acvp-ext-00 - I like the suggested rephrases. I am also going to delete the old files for the same specs to avoid confusion with largely duplicated content, coming soon in another PR.

artifacts/draft-celi-block-ciph-00.txt
```diff
@@ -578,7 +578,123 @@ Internet-Draft                Sym Alg JSON                 February 2018
...
+
+3.2.4.  AES Monte Carlo Test - CFB1 mode
+
+3.2.5.  AES Monte Carlo Test - CFB8 mode
+
+3.2.6.  AES Monte Carlo Test - CFB128 mode
+
+3.2.7.  TDES Monte Carlo Test - TECB mode
+
+3.2.8.  TDES Monte Carlo Test - TC8C mode
```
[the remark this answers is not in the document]
celic: :+1: As well, the title of the specification is "block cipher" but they are referred to as "symmetric cipher" through the specification. I know they refer to the same thing here, but it might be worthwhile to just change the name of the spec to match.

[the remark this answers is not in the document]
atvassilev: @celic, just refreshed my emacs skills of search and replace 👍. FYI, got a new laptop (MacBook Pro) last week and switched my tool set entirely to shell and emacs. Reminds me of my graduate school years when I was doing a lot of emacs, latex and gcc. Have to say, it feels a bit retro but nice.    

I'll commit shortly and after that the spec is yours as the new editor.

PR #431, opened 2018-08-27
Add EdDSA Specification

celic:
Adds an initial EdDSA spec for FIPS 186-5 (not yet officially released). 

This specification is versioned for v0.5. The goal is to have all modes of EdDSA available once v0.5 becomes available. 

In terms of implementation details, while the specification is not officially released, EdDSA will only be available on the demo server; not on the production server. The intention is to allow implementations to be tested on demo and ready for the official release to prod when the specification is released.

### DIFF
--- a/artifacts/acvp_sub_eddsa.html
+++ b/artifacts/acvp_sub_eddsa.html
@@ -973,7 +973,7 @@
 </tr>
 <tr>
 <td class="left">testType</td>
-<td class="left">The testType for the group</td>
+<td class="left">The testType for the group. Either AFT or BFT.</td>
 <td class="left">value</td>
 <td class="left">No</td>
 </tr>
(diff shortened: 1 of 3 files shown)


celic:
Also this was formatted for the "old" style of specifications (because it was very similar to ECDSA). It is missing the sections the new symmetric spec introduced, but can be updated later to reflect these.

smuellerDD:
Very nice that Curve25519 comes along!

May I inquire about the plans or even status on the ECDH version of the same mechanism? Note, EdDSA is the signature mechanism whereas with a slight modification you get the ECDH like implementation. The EdDSA is not too much in use as far as I can see, but the ECDH approach is in quite some use. Thanks

celic:
@smuellerDD I am not familiar with NIST's standardization work on ECDH. I can ask around and see if it is something we should implement alongside EdDSA.

atvassilev:
@smuellerDD We are definitely considering using these curves for ECDH. The question is not if but when. Depending on available resources and competing priorities, it could be the next version of 800-56A or some other standard. We (the 56-series team) are talking and will be deciding soon.

PR #442, opened 2018-09-18
Change kdf108 deferred to infoGeneratedByServer

### LINKED ISSUE #440: KDF108 Spec: Missing "deferred" in table
Hello,

I have a few things to discuss.

1) In Table 6, the "deferred" JSON appears to be missing.

2) In Table 6, it might be good to explicitly include a new element "ivLength" for the iv in bits, or indicate as to how the length can be implicitly inferred from another element (i.e. macMode). To make sure I'm understanding correctly, if the macMode is HMAC-SHA1 then the length of "iv" in bits would match the HMAC output length (160 in this case).

3) In Table 6, it would be great to also have an explicit length similar to point 2 above, or explanation how to calculate the length. Looking through some server JSON, it does not appear to be based on the macMode

4) In Table 9, it may be beneficial to give some more details on the expected fixedData. For instance, does it have a min/max length? Maybe point the reader to 800-108 document? I have read through 800-108 document and don't have a clear idea on what is expected from fixedData. An explanation, if only in this issue, would help me personally :)

Particularly for 2 and 3, knowing the expected length will help libraries that want to perform string length checks.

Hope that all made sense. If I'm misunderstanding anything please let me know.

Thanks!
Philip

fliphil:
Adds more clarity around lengths.
Added keyIn and iv to vector set response table.

Fixes #440

### DIFF
--- a/artifacts/acvp_sub_kdf108.html
+++ b/artifacts/acvp_sub_kdf108.html
@@ -957,8 +957,8 @@
 <td class="left"></td>
 </tr>
 <tr>
-<td class="left">infoGeneratedByServer</td>
-<td class="left">This flag indicates that the server is responsible for generating inputs (keyIn, iv). If "false", the client must generate the relevant inputs and attach to associated vector set response.</td>
+<td class="left">deferred</td>
+<td class="left">This flag indicates that the client is responsible for generating inputs (keyIn, iv). If "true", the client must generate the relevant inputs and attach to associated vector set response.</td>
 <td class="left">boolean</td>
 <td class="left">No</td>
 </tr>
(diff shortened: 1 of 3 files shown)


elliemdaw:
just want to make sure this change is on @hbooth or @celic's radar since this changes the name of an attribute, not just spec clarity :) so that we can keep the client and the server on the same page!

atvassilev:
Just labeled #440 accordingly.

elliemdaw:
Thank you @atvassilev !!

fliphil:
Since I added the "fixes" comment, when this PR is merged it will close that issue. Is it okay if that issue is closed, or should it stay open?

Thanks!
Philip

celic:
I don't believe this change is necessary. As I mentioned in #440, `deferred` is consistent and is used in all other algorithms that exhibit the same behavior. Even in RSA where the client registers the boolean `infoGeneratedByServer`, the server sends back a test case with the `deferred` property. 

If this wasn't clear from the specification, then that is a issue that needs to be resolved.

PR #449, opened 2018-10-02
Updates to introduction, architecture, authentication, messaging, mes…

bfussell:
…sage formats and references

### DIFF
--- a/src/draft-fussell-acvp-spec-00.xml
+++ b/src/draft-fussell-acvp-spec-00.xml
@@ -2,6 +2,8 @@
 <!DOCTYPE rfc SYSTEM "rfc2629.dtd" [
 <!ENTITY RFC2119 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.2119.xml">
 <!ENTITY RFC3339 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.3339.xml">
+<!ENTITY RFC5246 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.5246.xml">
+<!ENTITY RFC7230 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.7230.xml">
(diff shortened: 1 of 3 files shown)


src/draft-fussell-acvp-spec-00.xml
```diff
@@ -236,7 +233,10 @@
           </list>
         </t>
         <t>With that in mind there are several expectations when building a server used as a
-          validation authority. A validation authority shall use HTTPS, TLS 1.2 or greater and
+          validation authority. A validation authority shall use HTTPS
```
celic: RFCs use "SHALL" in capital letters to link them to RFC 2119 (like in section 2)

celic: This probably occurs in other locations... might make sense to do a 'find' for the words that appear in section 2.

src/draft-fussell-acvp-spec-00.xml
```diff
@@ -250,8 +250,32 @@
...
+              processes test responses.</t>
+            <t> ACV Proxy - Resides between the ACV server and ACV client to proxy the connection
+              for the client. This is particularly useful when the client does not support TLS, key
+              management or have signature capabilities and they are required by the server. An
+              example architecture is provided in Figure 3.</t>
+            <t> Device Under Test - Contains the crypto module under test which
+              can include various algorithms and functions that encrypt/decrypt, generate keys,
+              signatures, perform varifications and DRBG functions. May also contain the ACVP
+              client.</t>
+            <t> Crpyographic Module API - This is the interface, manual or otherwise,
```
celic: Cryptographic

src/draft-fussell-acvp-spec-00.xml
```diff
@@ -783,6 +778,70 @@
           </section>
       </section>
 
+
+      <section anchor="flow" title="Message Flow">
```
celic: I think this is a section people will jump to frequently to get off the ground quickly. Might be worth-while to make it its own section, or point to it from somewhere in the beginning of the document.

PR #450, opened 2018-10-11
Readable symmetric spec

celic:


celic:
@atvassilev I made the changes and uploaded a new draft.

Issue #237, opened 2018-03-02
SHA response submission fails


smuellerDD:
The SHA response submission fails for every type of SHA. The server answers:
< HTTP/1.1 100 Continue
< HTTP/1.1 400 Bad Request
< Server: Apache-Coyote/1.1
< Content-Type: application/json;charset=ISO-8859-1
< Content-Length: 131
< Date: Fri, 02 Mar 2018 17:04:31 GMT
< Connection: close

As snippet of the JSON for SHA-1:
```
  {
    "acvVersion":"0.4"
  },
  {
    "vsId":687,
    "testResults":[
      {
        "tcId":1,
        "md":"da39a3ee5e6b4b0d3255bfef95601890afd80709"
      },
      {
        "tcId":2,
        "md":"999a0a8cad102eb519ca788d6e0cd48ee9e6ca3e"
      },
```

and

```
     {
        "tcId":130,
        "resultsArray":[
          {
            "count":"0",
            "md":"3a234df4571bcc45baff3712ca288ee3c7eba43b"
          },
          {
            "count":"1",
            "md":"35bc8fd47a7bebc957de7154164edd650fa2d704"
          },
          {
            "count":"2",
            "md":"02124b597645ef7b15de4386dc31ae7ff77f18e9"
          },
```

celic:
Is count supposed to be a string or an integer?

smuellerDD:
I just reviewed the latest spec and the count value is gone entirely. When using the following JSON for the MCT, the server accepts it:

```
      {
        "tcId":130,
        "resultsArray":[
          {
            "md":"3a234df4571bcc45baff3712ca288ee3c7eba43b"
          },
          {
            "md":"35bc8fd47a7bebc957de7154164edd650fa2d704"
          },
```
However, the server responds: with the error: "reason" : "ResultsArray did not contain expected element Message" when fetching the result.


Note, the following example in the SHA spec does not seem to be valid JSON as an array does not contain key-value pairs:
```
   [
     { "acvVersion": "0.4" },
     { "vsId": 1564,
       "testResults": [
       {
           "tcId": 2175,
           "resultsArray": [
               "md": "23d1f8655a7c9c2bc3bb56868483611e67cb1b3a",
               "md": "d3d672efb8e569dc7a37c991a9fc51282412196f"
           ]
       }]
     }]
```

smuellerDD:
[...]

PR #489, opened 2018-11-07
Hash revisions

celic:


celic:
Need to check on #237 to ensure it is fixed before merging.

Issue #469, opened 2018-11-02
v0.5: AES CFB: ptLen / ctLen


smuellerDD:
The symmetric cipher sipec does not contain the definition of the keywords of ptLen and ctLen any more that used to be present for CFB. The AES-CFB1/8 vectors do not contain the keyword nor do the responses need that keyword.

Though, the examples in the spec still contain ptLen/ctLen. I guess they would need to be removed?

Kritner:
@smuellerDD I saw you commented on the PR, so I'm closing this issue.  Let me know if something comes up!

PR #497, opened 2018-11-08
Symmetric spec `dataLen`, `ptLen`, `ctLen` -> `payloadLen`

Kritner:
Updates symmetric spec to use `payloadLen` as the property name, taking place of `ptLen`, `ctLen`, and `dataLen`

What is everyone's thoughts on this? I know we're kind of late in the v0.5 game here, but right now the spec isn't matching the server, and `dataLen` can be somewhat ambiguous (IMO).


Kritner:
Note this is not done yet, i just wanted to get something out there for visibility, i have a few more commits to make

Kritner:
Note - these changes are now completed, and will be deployed to the demo server in the next release.  I will merge this PR into master at that point.

Kritner:
This will close #469 once merged in

smuellerDD:
confirmed, working

PR #515, opened 2018-11-14
Feature/dsa iana

Kritner:
Additional (remaining?) IANA updates for DSA
* Added test type information
* Added requirements covered information


Kritner:
I believe this is mostly done - unless there's something obviously wrong, should probably be merged in prior to the changes discussion on the call this afternoon regarding IANA

PR #532, opened 2018-12-17
Large file submission doc

dstroud3:
Large file submission documentation


bfussell:
Looks good, but we will need for all 3 files to be updated with the new text. By that I mean in addition to the .xml file, the .html and .txt files will need to be added to the PR.

dstroud3:
Thanks for the feedback, I'll get on that!

smuellerDD:
I am not sure I can chime in here and what discussions already happened. But the current specification does not comply with the description of all other RESTAPI endpoints. I.e. the table 5.2 needs to be updated as Apostol mentioned. Also, chapter 11 needs to be updated with:

- which endpoint returns what exact data indicating that the large support is present

- how exactly is the large endpoint to be used (which JSON data is to be sent, how shall the existing data be segmented if the large endpoint has insufficient limit (see the GCM tests: the GCM JSON results are some 20 MB in size - how shall that be segmented?))

- how is the submitted data combined with the vsID it is intended for?

Without such information, I would not know how to implement the large support.

Thanks

dstroud3:
Not a problem, I will make the required changes and include your feedback.

atvassilev:
@smuellerDD, thanks for the feedback, you are always welcome to chime in. On another note, the ICMC2019 organizers asked us for a tutorial session on ACVP, on the day before the first. Would you be available and interested in joining us for this? I do not know the exact details of the show and tell that is being planned but wanted to know if you are interested.

smuellerDD:
Am Dienstag, 18. Dezember 2018, 18:07:33 CET schrieb Apostol Vassilev:

Hi Apostol,

> @smuellerDD, thanks for the feedback, you are always welcome to chime in. On
> another note, the ICMC2019 organizers asked us for a tutorial session on
> ACVP, on the day before the first. Would you be available and interested in
> joining us for this? I do not know the exact details of the show and tell
> that is being planned but wanted to know if you are interested.

Sure, I am interested and would be ready if you want to have a 2nd 
implementation demonstration in addition to libacvp. As I have quite a number 
of crypto libs available as well as an iOS app with a GUI that provides some 
ready-to-interpret output, I can also provide additional demonstration if 
requested.

Ciao
Stephan

atvassilev:
Stephan,

OK great, I will pass this along to the group here and we will be in touch.

smuellerDD:
Is the documented behavior already active?

When I examine the login, I do not see "largeEndpointRequired" nor "sizeConstraintMessage".

In the spec, please add a new section to chapter 10 with the very same structure as all other sections in chapter 10. Note, the current documentation misses the keyword holding the large-URI returned by the POST request.

dstroud3:
The documented behavior is live in the development environment right now.  We will run one final test tomorrow in development to make sure everything is working as expected with a large submission.

smuellerDD:
I was not referring to the picture in 10.4.

My issue is that from the SPEC, I do not know how the answer from the ACVP looks like that contains the URI that I should use. Is it JSON? If yes, which keyword shall I search for?

If the functionality is online, I must be misreading the spec. The spec says that when using /login, I get the JWT auth token *and* additionally the maximum size of a request. I get the token but I do not get the maximum size in the ACVP server answer.

smuellerDD:
I just ran another complete session to download a test vector. None of the JSON messages contain either key word of largeEndpointRequired or sizeConstraintMessage.

celic:
@smuellerDD this is not yet available in the public environment. @dstroud3 meant to say this is only being tested in the private development environment right now.

smuellerDD:
Wtih the latest update, it seems the functionality is now live. But I am not sure how to interpret the following:

```
{"largeEndpointRequired":false},{"sizeConstraintMessage":null}
```

I guess the sizeConstraintMessage of NULL looks like an error on the ACVP server side.

But what does largeEndpointRequired==false really mean? Does that keyword mean the server does not support the large endpoint? Or is it the case that it is not mandatory to use the large endpoint. If it is not mandatory, does it mean that it is even not mandatory for data sizes above the threshold?

smuellerDD:
I am trying to implement the parsing side of the ACVP server response and found that the JSON structure is not really helpful.

Currently, the server response looks like:

```
[
   {"acvVersion":"0.5"},
   {"accessToken":"..."},
   {"largeEndpointRequired":false},
   {"sizeConstraintMessage":null}]
```

I.e. we have one array with individual JSON objects. I think this structure is wrong as follows: In all other responses, the ACVP version is the first array entry (like here). All data is provided with the 2nd array entry. Here, however, we have additional data with the 3rd and 4th array entry. As arrays are ordered lists, I think this approach should be changed back to comply with all other ACVP responses. Thus, the information about the large endpoint should be part of the accessToken JSON object as follows:

```
[
   {"acvVersion":"0.5"},
   {
     "accessToken":"...",
     "largeEndpointRequired":false,
     "sizeConstraintMessage":null
   }
```

dstroud3:
Login response has been adjusted to fit the comment

hbooth:
trying to pick this up and get this pushed into master.
I merged master back into this branch to resolve the merge conflicts, then reviewed the changes and made a number of updates to the documentation. Please review and let me know what needs to be updated. Otherwise please accept this pull request and merge into master.

smuellerDD:
The changes seem to contain a wrong response in the section 12.2: shouldn't there be a URL returned by the ACVP server?

Otherwise, the changes look good and I see the large-support in the login responses as requested. Though it seems that the large-support is not enabled on the ACVP server since largeEndpointRequired is false.

hbooth:
I have been looking at this (and the implementation) and have noticed a number of deficiencies. I will be updating this PR a little later today.

hbooth:
fixed the response example and added in the accessToken property.

atvassilev:
@hbooth, are you done with the updates you planned for this round? I looked at the diff and seems OK. Let me know if you plan to work on it some more.

hbooth:
I believe the spec is pretty close to where it needs to be now. Still working on finalizing our implementation, but I don't expect to need to make anymore spec changes.

Issue #545, opened 2019-02-07
SHAKE "outputLength" registration JSON rejected


fliphil:
Here's the error...

`{
  "acvVersion" : "0.5",
  "error" : "Incorrectly formatted JSON (1:642): expected array value and detected scalar: outputLength"
}`

**Algorithm registration**
`[{"acvVersion":"0.5"},{"algorithms":[{"algorithm":"SHA-1","inBit":false,"inEmpty":true},{"algorithm":"SHA2-224","inBit":false,"inEmpty":true},{"algorithm":"SHA2-256","inBit":false,"inEmpty":true},{"algorithm":"SHA2-384","inBit":false,"inEmpty":true},{"algorithm":"SHA2-512","inBit":false,"inEmpty":true},{"algorithm":"SHA3-224","inBit":false,"inEmpty":true},{"algorithm":"SHA3-256","inBit":false,"inEmpty":true},{"algorithm":"SHA3-384","inBit":false,"inEmpty":true},{"algorithm":"SHA3-512","inBit":false,"inEmpty":true},{"algorithm":"SHAKE-128","inBit":false,"inEmpty":true,"outBit":false,"outputLength":{"min":16,"max":65536,"increment":8}},{"algorithm":"SHAKE-256","inBit":false,"inEmpty":true,"outBit":false,"outputLength":{"min":16,"max":65536,"increment":8}}]}]`

**Endpoint in which the error is experienced**
https://demo.acvts.nist.gov:443/acvp/acvp/v1/testSessions| POST

**Expected behavior**
It should accept the "outputLength" according to this document: https://github.com/usnistgov/ACVP/blob/master/artifacts/draft-celi-sha3-00.txt

Rasstrelnikov:
I would guess this is because your value for outputLength is not an array. A domain is an array that contains ranges (which is the object you have) or integers. So I believe
 ```"outputLength":[{"min":16,"max":65536,"increment":8}]```

celic:
@acronymhaterheidi is correct. A domain could be an array of individual values, mixed with these mentioned range objects.

fliphil:
Okay thanks. I was confused because the example in that spec document doesn't show an array. I think the document Appendix A. needs to be updated to the correct array value.

celic:
If the spec is incorrect here, then it needs to be corrected. Feel free to submit a PR if you are interested or have the opportunity.

Kritner:
This seems to be resolved with @acronymhaterheidi's PR in #547

PR #547, opened 2019-02-08
Change outLength to domain in SHAKE registration example

acronymhaterheidi:
Fixes error pointed out in #545

### DIFF
--- a/src/draft-celi-sha3-00.xml
+++ b/src/draft-celi-sha3-00.xml
@@ -465,10 +465,11 @@
 	"inBit": true,
 	"inEmpty": true,
 	"outBit": true,
-	"outLength": {
+	"outLength": [
+        {
 		"min": 16,
 		"max": 1024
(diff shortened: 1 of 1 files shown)


celic:
This does not contain the compiled artifacts (txt, html) but does fix the issue in the xml. (The compiled artifacts arguably don't belong in the repository at all). I'm fine approving this and merging it in and adding the artifacts later. Thoughts @atvassilev ?

bfussell:
I guess the other 2 files, .txt and .html, need updating as well ?

celic:
I incorporated this change into #551 and included the compiled artifacts.

celic:
Reopening because the other PR #551 is not being merged at the moment. This should be merged to fix the listed issue. Sorry for the confusion.

celic:
I will merge this and follow up with the compiled artifacts.

PR #551, opened 2019-02-13
IANA SHA update

celic:


atvassilev:
@celic I think this should wait for after the IETF. We talked about only the symmetric block ciphers. I restricted the IANA registry to only the block cipher algorithms - see PR #552

celic:
Agreed. It can wait.

Issue #543, opened 2019-02-02
AES XPN Confusion


acronymhaterheidi:
The block cipher spec includes this example registration:
 

                    "algorithm": "AES-XPN",
                    "prereqVals" : [{"algorithm" : "AES", "valValue" : "123456"},
                            {"algorithm" : "DRBG", "valValue" : "123456"}],
                    "direction": [
                        "encrypt",
                        "decrypt"
                    ],
                    "ivGen": "internal",
                    "ivGenMode": "8.2.2",
                    "saltGen": "internal",
                    "keyLen": [
                        128,
                        192,
                        256
                    ],
                    "tagLen": [
                        96,
                       128
                    ],
                   "payloadLen": [
                        0,
                        128
                    ],
                   "aadLen": [
                        120,
                        128
                   ]

This contradicts the rest of the spec in 2 ways that I can see:

1. There is no ivLen value. This may not be required, but I interpreted "The following grid outlines which properties are REQUIRED, as well as the possible values a server MAY support for each authenticated block cipher algorithm:" as meaning that if there is a value in Table 4, it is required for the algorithm to have that property. 

2. The payloadLen has 0 as a value. I was not entirely sure what {"Min":1, "Max":xxx, "Inc": any} means but I interpreted this as allowing any strictly positive integer.

Kritner:
AES-XPN only supports IVLengths of 96 bits. Payload length is a domain, described here: https://usnistgov.github.io/ACVP/artifacts/draft-fussell-acvp-spec-00.html#rfc.section.14.4

Rasstrelnikov:
I am aware that the Payload length is a domain. What I suppose I mean to say is that the payload length in the example does not seem to be a valid length. In the Authenticated Block Ciphers Applicability Grid it lists {"Min":1, "Max":xxx, "Inc": any} for the payloadLen for AES-XPN. Assuming that the "xxx" in "Max" means that there is no limit, 0 would not be a valid length because it is below the minimum of 1. 128 would be a valid length. 

If ivLen can only be 96 bits, why is the entry in the table {"Min": 8, "Max": 1024, "Inc": any}? And does that mean that it is not required? My reading of the table as it is is that the ivLen is required and that it can be any bit length between 8 and 1024, inclusive.

celic:
Sorry... "xxx" here means it will be filled in later. That later happens to be this week.

celic:
This has been updated in #558 by the way.

[...]

PR #558, opened 2019-02-14
Adds "revision" to Symmetric spec

celic:
Also updates examples so there are no extremely long strings. This is a requirement for IETF publications.


celic:
Also fixes #543

acronymhaterheidi:
This replaces `payloadLen` with `dataLen` in some of the tables but nowhere else. Is this a new key for some of the block cipher modes?

celic:
Thank you for the heads up. This is unintentional as I pasted my tables from an excel sheet. The excel sheet wasn't updated with the proper names for those properties. I'll update this PR.

celic:
I've got some final updates for this coming today.

PR #563, opened 2019-02-19
Add message length to SHA1,2

celic:
This change removes the two boolean values (`"inEmpty"` and `"inBit"`) from the registrations for SHA1 and SHA2 in favor of a domain that expresses the same information (and more).

### DIFF
--- a/src/draft-celi-sha-00.xml
+++ b/src/draft-celi-sha-00.xml
@@ -234,7 +234,7 @@ MD[0] = MD[1] = MD[2] = SEED
 					<c>string</c>
 
 					<c>messageLength</c>
-					<c>The message lengths supported by the IUT. Minimum allowed is 0, maximum allowed is 65535.</c>
+					<c>The message lengths in bits supported by the IUT. Minimum allowed is 0, maximum allowed is 65535.</c>
 					<c>domain</c>
 				</texttable>
 				<t>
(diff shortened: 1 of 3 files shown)


celic:
The intention is for this to go live with the server as of v0.5.6 (likely next week, week of 2/25/19).

src/draft-celi-sha-00.xml
```diff
@@ -233,13 +233,9 @@ MD[0] = MD[1] = MD[2] = SEED
...
 
-					<c>inBit</c>
-					<c>Implementation does accept bit-oriented messages</c>
-					<c>boolean</c>
-
-					<c>inEmpty</c>
-					<c>Implementation does accept null (zero-length) messages</c>
-					<c>boolean</c>
+					<c>messageLength</c>
+					<c>The message lengths supported by the IUT. Minimum allowed is 0, maximum allowed is 65535.</c>
```
bfussell: Do we need to specify the lengths are in bits here ?

atvassilev: I agree, the text should state the expected lengths are in bits.

smuellerDD:
While you are chainging the SHA spec, may I propose to remove the Msg field in the MCT responses? Note, this field is not required for a verification.

PS: I ask from the following perspective: For ICMC, I will hold a presentation showing some pitfalls when using ACVP. One of the topics is whether it is possible to have an ACVP2CAVS and CAVS2ACVP converter. I already built one. This work shows that SHA MCT CAVS responses cannot be converted into ACVP responses, because CAVS lacks the Msg information. As I do not see the necessity to have the Msg field, I would like to have it removed to allow users to implement a CAVS to ACVP converter.

celic:
@smuellerDD would you mind opening up a separate issue ticket for this? I think it's a good idea. MCT msg was only present for debugging but the message is trivial to construct anyways. It doesn't necessarily have a place in an automated system. (And frankly, 99 of the digests don't have a place either but I'd keep that as a separate task).

celic:
Updates to include that the length is in bits.

smuellerDD:
Will do, please disregard my note as part of this issue then.

Re limiting the number of digests: I am all for it.

Ciao
Stephan

—
Cui bono?

Issue #504, opened 2018-11-09
Need test spec revision in addition to protocol spec version


atvassilev:
The IANA registry for ACVP introduced the concept of revisions for test specifications - see PR #499. This means that going forward we would be able to improve testing and manage the corresponding evolution from old to new independently of the protocol itself. Currently, the version of the test specification for a given algorithm is the same as the version of the protocol and on top of this, all test specifications are versioned at the same level all the time. 

We need to introduce a revision property in the specification for each algorithm test and manage it independently from the version of the protocol, as shown in the example below. Every algorithm that is included in the registration should have its own revision number. Similarly, the test vectors and the responses will contain the same revision property.  

```json
           {
                    "algorithm": "AES-GCM",
                    **"revision": "1.0",**
                    "prereqVals" : [{"algorithm" : "AES", "valValue" : "123456"},
                            {"algorithm" : "DRBG", "valValue" : "123456"}],
                    "direction": [
                        "encrypt",
                        "decrypt"
                    ],
                    "ivGen": "internal",
                    "ivGenMode": "8.2.2",
                    "keyLen": [
                        128,
                        192,
                        256
                    ],   
```

Kritner:
This has been deployed to demo.  Let us know if there are any issues.

PR #602, opened 2019-04-04
Testing Revision

Kritner:
A new capability registration property for each algorithm that controls which "testing revision" the vector generation/validation should utilize.

As a starting point, all algorithms will be at testing revision "1.0", but as time goes on, as testing changes, those testing revisions will theoretically change independent of the ACVP protocol.

This `revision` property will be incremented on a per algorithm basis - so "AES-GCM" could be at testing revision "1.5", while "SHA2" remains at "1.0", while still being separate from the ACVP protocol version of "2.0".

### DIFF
--- a/artifacts/acvp_sub_drbg.html
+++ b/artifacts/acvp_sub_drbg.html
@@ -400,12 +400,12 @@
 <link href="#rfc.authors" rel="Chapter">
 
 
-  <meta name="generator" content="xml2rfc version 2.6.2 - http://tools.ietf.org/tools/xml2rfc" />
+  <meta name="generator" content="xml2rfc version 2.21.1 - https://tools.ietf.org/tools/xml2rfc" />
   <link rel="schema.dct" href="http://purl.org/dc/terms/" />
 
   <meta name="dct.creator" content="Vassilev, A., Ed." />
   <meta name="dct.identifier" content="urn:ietf:id:draft-ietf-acvp-subdrbg-0.5" />
-  <meta name="dct.issued" scheme="ISO8601" content="2018-11-13" />
+  <meta name="dct.issued" scheme="ISO8601" content="2018-11-01" />
   <meta name="dct.abstract" content="This document defines the JSON schema for testing DRBG implementations from SP 800-90A  with the ACVP specification." />
   <meta name="description" content="This document defines the JSON schema for testing DRBG implementations from SP 800-90A  with the ACVP specification." />
 
@@ -426,10 +426,10 @@
 </tr>
 <tr>
 <td class="left">Intended status: Informational</td>
-<td class="right">November 13, 2018</td>
+<td class="right">November 1, 2018</td>
 </tr>
 <tr>
-<td class="left">Expires: May 17, 2019</td>
+<td class="left">Expires: May 5, 2019</td>
 <td class="right"></td>
 </tr>
 
@@ -444,12 +444,12 @@
 <p>This document defines the JSON schema for testing DRBG implementations from SP 800-90A  with the ACVP specification.</p>
 <h1 id="rfc.status"><a href="#rfc.status">Status of This Memo</a></h1>
 <p>This Internet-Draft is submitted in full conformance with the provisions of BCP 78 and BCP 79.</p>
-<p>Internet-Drafts are working documents of the Internet Engineering Task Force (IETF).  Note that other groups may also distribute working documents as Internet-Drafts.  The list of current Internet-Drafts is at http://datatracker.ietf.org/drafts/current/.</p>
+<p>Internet-Drafts are working documents of the Internet Engineering Task Force (IETF).  Note that other groups may also distribute working documents as Internet-Drafts.  The list of current Internet-Drafts is at https://datatracker.ietf.org/drafts/current/.</p>
 <p>Internet-Drafts are draft documents valid for a maximum of six months and may be updated, replaced, or obsoleted by other documents at any time.  It is inappropriate to use Internet-Drafts as reference material or to cite them other than as "work in progress."</p>
-<p>This Internet-Draft will expire on May 17, 2019.</p>
+<p>This Internet-Draft will expire on May 5, 2019.</p>
 <h1 id="rfc.copyrightnotice"><a href="#rfc.copyrightnotice">Copyright Notice</a></h1>
 <p>Copyright (c) 2018 IETF Trust and the persons identified as the document authors.  All rights reserved.</p>
-<p>This document is subject to BCP 78 and the IETF Trust's Legal Provisions Relating to IETF Documents (http://trustee.ietf.org/license-info) in effect on the date of publication of this document.  Please review these documents carefully, as they describe your rights and restrictions with respect to this document.  Code Components extracted from this document must include Simplified BSD License text as described in Section 4.e of the Trust Legal Provisions and are provided without warranty as described in the Simplified BSD License.</p>
+<p>This document is subject to BCP 78 and the IETF Trust's Legal Provisions Relating to IETF Documents (https://trustee.ietf.org/license-info) in effect on the date of publication of this document.  Please review these documents carefully, as they describe your rights and restrictions with respect to this document.  Code Components extracted from this document must include Simplified BSD License text as described in Section 4.e of the Trust Legal Provisions and are provided without warranty as described in the Simplified BSD License.</p>
 
   
   <hr class="noprint" />
@@ -1025,6 +1025,20 @@
 <td class="left"></td>
 </tr>
 <tr>
+<td class="left">revision</td>
+<td class="left">The algorithm testing revision to use.</td>
+<td class="left">value</td>
+<td class="left">1.0</td>
+<td class="left">No</td>
+</tr>
+<tr>
+<td class="left"></td>
+<td class="left"></td>
+<td class="left"></td>
+<td class="left"></td>
+<td class="left"></td>
+</tr>
+<tr>
 <td class="left">derFuncEnabled</td>
 <td class="left">derivation function option, see the notes below            <a href="#caps_table" class="xref">Table 5</a> for more information.</td>
 <td class="left">boolean</td>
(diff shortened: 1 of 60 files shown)


smuellerDD:
Which version is it going to be right now? I.e. how shall we go about the current transition phase?

Kritner:
All algorithms are going to be starting at `"revision": "1.0"`, it's only changes to testing that occur after their initial release that will bump the testing revision.

This change is not currently on demo, but AFAIK should be coming shortly.  I'm in the process of getting all the specs and samples up to date with the `revision` description, valid values (currently only 1.0), and json sample updates.

The expectation is that `revision` will be provided in the capability registration, the ACVP server will send it back through in the prompt file, and it is ***not*** expected back in the response file.  So hopefully this will be a minimal amount of work from the client's perspective, especially considering there will be no need for "special logic" (at least currently, since there is only testing revision 1.0) to roll out this feature.

The overall intention of the revision feature is that testing may/will change over time - if we were to change the MCT test for revision "2.0" of a specific algorithm, it's pretty important to both the server and client to know which iteration of the MCT test to utilize (1.0 or 2.0 as example).  This testing revision will help with such logic switching, and will allow testing revisions on a per algorithm basis to be kept separate from the ACVP protocol itself.

smuellerDD:
Am Donnerstag, 4. April 2019, 14:33:04 CEST schrieb Russ Hammett:

Hi Russ,

> All algorithms are going to be starting at `"revision": "1.0"`, it's only
> changes to testing that occur after their initial release that will bump
> the testing revision.
> 
> This change is not currently on demo, but AFAIK should be coming shortly. 
> I'm in the process of getting all the specs and samples up to date with the
> `revision` description, valid values (currently only 1.0), and json sample
> updates.
> 
> The expectation is that `revision` will be provided in the capability
> registration, the ACVP server will send it back through in the prompt file,
> and it is ***not*** expected back in the response file.  So hopefully this
> will be a minimal amount of work from the client's perspective, especially
> considering there will be no need for "special logic" (at least currently,
> since there is only testing revision 1.0) to roll out this feature.

This makes sense and for me it is a minimal effort indeed. I was just 
wondering how it is added to the demo server considering that it can be 
disruptive. If you roll it out for all cipher definitions with one update, 
this would be perfect.
> 
> The overall intention of the revision feature is that testing may/will
> change over time - if we were to change the MCT test for revision "2.0" of
> a specific algorithm, it's pretty important to both the server and client
> to know which iteration of the MCT test to utilize (1.0 or 2.0 as example).
>  This testing revision will help with such logic switching, and will allow
> testing revisions on a per algorithm basis to be kept separate from the
> ACVP protocol itself.

But this will imply the revision version is added to the test vector JSON data 
the ACVP server returns?!


Ciao
Stephan

Kritner:
> All algorithms are going to be starting at `"revision": "1.0"`, it's only
 changes to testing that occur after their initial release that will bump
 the testing revision.

> This change is not currently on demo, but AFAIK should be coming shortly.
 I'm in the process of getting all the specs and samples up to date with the
 `revision` description, valid values (currently only 1.0), and json sample
 updates.

> The expectation is that `revision` will be provided in the capability
 registration, the ACVP server will send it back through in the prompt file,
 and it is ***not*** expected back in the response file.  So hopefully this
 will be a minimal amount of work from the client's perspective, especially
 considering there will be no need for "special logic" (at least currently,
 since there is only testing revision 1.0) to roll out this feature.

>> This makes sense and for me it is a minimal effort indeed. I was just
wondering how it is added to the demo server considering that it can be
disruptive. If you roll it out for all cipher definitions with one update,
this would be perfect.

Yes, this will be rolled out to all algorithms simultaneously in the next demo push.

> The overall intention of the revision feature is that testing may/will
 change over time - if we were to change the MCT test for revision "2.0" of
 a specific algorithm, it's pretty important to both the server and client
 to know which iteration of the MCT test to utilize (1.0 or 2.0 as example).
  This testing revision will help with such logic switching, and will allow
 testing revisions on a per algorithm basis to be kept separate from the
 ACVP protocol itself.
>> But this will imply the revision version is added to the test vector JSON data
the ACVP server returns?!

Sorry, I'm not understanding the concern here.  If (as an example) the MCT logic were different from a generation/validation perspective in between revisions "1.0" and "2.0", then both the ACVP Server and IUT would need to know which "revision" of the MCT test to utilize for generation, validation, and dispatching to the module under test.

From a client perspective, the revision will be included in the capabilities registration, and sent back to the client via the prompt file.  This step of sending the revision back in the prompt file ***may*** be unnecessary since the client could theoretically keep all of this information associated to the test vector ID provided back by the ACVP server; however we are already including the "algorithm/mode" with the prompt file, it seems to make sense to include the revision as well.

hbooth:
To chime in here the best example of when this would/will be used is when a new revision of an algorithm such as the transition from FIPS 186-2 to FIPS 186-4 (or a future FIPS 186-5) is put into place. We will need to be able to describe and differentiate between the testing of those two standards.

The initial example Russ is describing is also likely. If we find we need to improve our testing and those improvements result in a modification to what is needed in the testing inputs and outputs then we will need to update the revision in that case too. In all cases, an implementation is going to need to understand beforehand what is necessary to successfully pass testing.

Adding `revision` allows us to handle these types of changes independent of the the ACV protocol version moving forward.

In essence the addition of the `revision` property is adding another tuple to the composite key that describes the algorithm testing: `name, mode, revision`.

atvassilev:
To add to @hbooth's example, another case that shows the need for revisions is the transition between FIPS 186-4 and FIPS 186-5 (currently in draft, to be published this year). The method for generating random probable primes is changing and accordingly the method for testing compliance to that method of generating primes is going to be substantially different. In FIPS 186-4 the IUT only provides numbers that the server verifies that they are prime. In FIPS 186-5, we are asking the IUT to seed and use a DRBG and draw random values from it. The server will ask the IUT to provide the seed to the DRBG so that it can reproduce fully the method for generating the prime numbers computed by the IUT. The intent here is to prevent special post-processing of the random data that may lead to vulnerabilities similar to ROCA.

smuellerDD:
Am Donnerstag, 4. April 2019, 15:07:01 CEST schrieb Russ Hammett:

Hi Russ,

> >> But this will imply the revision version is added to the test vector JSON
> >> data
> the ACVP server returns?!
> 
> Sorry, I'm not understanding the concern here.  If (as an example) the MCT
> logic were different from a generation/validation perspective in between
> revisions "1.0" and "2.0", then both the ACVP Server and IUT would need to
> know which "revision" of the MCT test to utilize for generation,
> validation, and dispatching to the module under test.
> 
> From a client perspective, the revision will be included in the capabilities
> registration, and sent back to the client via the prompt file.  This step
> of sending the revision back in the prompt file ***may*** be unnecessary
> since the client could theoretically keep all of this information
> associated to the test vector ID provided back by the ACVP server; however
> we are already including the "algorithm/mode" with the prompt file, it
> seems to make sense to include the revision as well.

Look at my implementation: I have a proxy that handles the entire ACVP 
communication. It registers with ACVP and, say, tells the revision. Then the 
ACVP proxy obtains the test vectors into an offline database.

That database is handed to the ACVP parser which has no relationship with the 
ACVP server. And it has therefore no idea about any ACVP sessions. I.e. it 
only acts on the JSON test vectors obtained from the server.

Thus, if the revision has an impact on the test definition like which MCT to 
run, the ACVP parser would need to know about it. And it would only know about 
it via the ACVP test vector definitions.

Hence, my question to have it part of the test vector received from the ACVP 
server.

To go even further, assume that the ACVP proxy is operated by our lab, but the 
ACVP parser is operated by a vendor. All that we do is to ship the test 
vectors and responses back and forth. As such, the vendor's parser has no idea 
about what the proxy "agreed" on with the ACVP server beyond the ACVP test 
vectors.


Ciao
Stephan

smuellerDD:
Am Donnerstag, 4. April 2019, 15:18:56 CEST schrieb Harold Booth:

Hi Harold,

> To chime in here the best example of when this would/will be used is when a
> new revision of an algorithm such as the transition from FIPS 186-2 to FIPS
> 186-4 (or a future FIPS 186-5) is put into place. We will need to be able
> to describe and differentiate between the testing of those two standards.

I am not disputing that and I am actively trying to support it. Yet, consider 
my (and even the ACVP spec's) scenario of having a proxy and a parser (IUT 
executor).
> 
> The initial example Russ is describing is also likely. If we find we need to
> improve our testing and those improvements result in a modification to what
> is needed in the testing inputs and outputs then we will need to update the
> revision in that case too. In all cases, an implementation is going to need
> to understand beforehand what is necessary to successfully pass testing.
> 
> Adding `revision` allows us to handle these types of changes independent of
> the the ACV protocol version moving forward.
> 
> In essence the addition of the `revision` property is adding another tuple
> to the composite key that describes the algorithm testing: `name, mode,
> revision`.

I agree fully here. All I am asking is to add the revision version into the 
test vector returned by the ACVP server. 


Ciao
Stephan

Kritner:
#606, closes #504

Kritner:
I think I got everything from the individual algorithms, I've unmarked WIP, but I would think keeping this PR open and merging once the revision changes go out to demo would be the least confusing?

smuellerDD:
Maybe I have misread the SPEC updates. I see, for example in the DRBG update that the test vectors do contain a revision.

Therefore, please disregard my comment.

Though I would like to raise another one. I would like to request the version to be a JSON integer instead of a string. With an integer, we can handle comparisons much easier. For example, you will use a higher version for, say FIPS 186-5 that requires some, say, changes in PKCS siggen (just for the sake of discussion. Now, at one point 186-6 comes along that requires changes in PSS siggen, i.e. the same test vector is affected, but different algos are modified.

Thus, for PKCS siggen I can simply say all ACVP vectors which revision > x or <y are covered with my implementation. Using a string that symbolizes some sort of number, this is not easily possible, especially when the numbering schema is not defined.

smuellerDD:
For the records: as discussed, no objections any more. I withdraw all my comment.s

PR #611, opened 2019-04-05
This commit has a purposeful error in the xml, should hopefully throw…

Kritner:
… an error in the build process


Kritner:
Nice 👍

Issue #614, opened 2019-04-06
SPEC: signature keyword still present


smuellerDD:
Section 11.14.4.1 still contains the signature in the example.

celic:
This is in the ACVP Specification itself? And this example is outside any algorithms correct?

hbooth:
Correct, this is something I need to fix...

hbooth:
PR #619 resolves this issue.

PR #619, opened 2019-04-08
removed signature from example per issue #614

hbooth:


celic:
There is no need to push the artifacts anymore.

PR #659, opened 2019-06-07
WIP: ANS x9.42 KDF

celic:
@acronymhaterheidi, This is not yet enabled on the server, but here is the spec with some examples so you can prepare. I recall you asked about it at ICMC. 

Trying something new here for document generation. I used [Metanorma](https://github.com/metanorma/metanorma) to generate the XML and then compiled that with the usual `xml2rfc`.

### DIFF
--- a/artifacts/draft-celi-acvp-ans-x942.html
+++ b/artifacts/draft-celi-acvp-ans-x942.html
@@ -382,8 +382,8 @@
 <link href="#rfc.section.5" rel="Chapter" title="5 Capabilities Registration">
 <link href="#rfc.section.5.1" rel="Chapter" title="5.1 Prerequisites">
 <link href="#rfc.section.5.2" rel="Chapter" title="5.2 Property Registration">
-<link href="#rfc.section.5.2.1" rel="Chapter" title="5.2.1 Valid Hash Functions">
 <link href="#rfc.section.5.3" rel="Chapter" title="5.3 Registration Example">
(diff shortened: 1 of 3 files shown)


acronymhaterheidi:
Thank you!

acronymhaterheidi:
The example registration does not seem to match the description of the properties- the "kdfType", "keyLen" and other properties described in the table in Section 5.2 are supposed to be inside the "capabilities" array per the table in Section 5.3. but in the example they are not.

celic:
You are correct. The properties should be flattened out, not inside a capabilities array. I'll correct that.

celic:
@acronymhaterheidi updated.

celic:
@acronymhaterheidi keep me posted on your testing of the algorithm. I struggled to find test vectors for the implementation of the KDFs. This is an algorithm not present in the old system so we had to write a new crypto implementation. If you have any issues getting good results with a "trusted" implementation, please let me know.

acronymhaterheidi:
@celic our vendor who has an implementation has decided to go with exclusively CAVS testing for their current validation to avoid redeveloping their test harness (which makes me a little sad only because I have developed a slight fixation on the testing for this KDF ¯\_(ツ)_/¯). I'll try to find an open source implementation and do some testing regardless.

celic:
That is their choice. They'll have to join ACVP eventually. I appreciate the effort.

PR #662, opened 2019-06-11
WIP: Adds PBKDF

celic:
@smuellerDD I know you were asking about these... Here is the specification for PBKDF.


smuellerDD:
First of all, thanks.

Apart from open areas, please allow me to make the following comments:

- iterationCount is limited to 10 to 1000 - We have seen (valid - as strange as it may sound) implementations of having a count of 1 and counts of some 10,000,000. Thus, may I ask for increasing the range?

- I am not sure that the saltLen lower end range of 128 bits is good - I think we should allow a lower value here

smuellerDD:
When requesting the following, the server returns an error:

```
[
  {
    "acvVersion":"1.0"
  },
  {
    "isSample":false,
    "operation":"register",
    "certificateRequest":"no",
    "debugRequest":"yes",
    "production":"no",
    "encryptAtRest":"yes",
    "algorithms":[
      {
        "revision":"1.0",
        "algorithm":"kdf-components",
        "mode":"pbkdf",
        "prereqVals":[
          {
            "algorithm":"SHA",
            "valValue":"same"
          }
        ],
        "iterationCount":[
          {
            "min":10,
            "max":1000,
            "increment":1
          }
        ],
        "keyLen":[
          {
            "min":8,
            "max":4096,
            "increment":8
          }
        ],
        "passwordLen":[
          {
            "min":8,
            "max":128,
            "increment":1
          }
        ],
        "saltLen":[
          {
            "min":128,
            "max":4096,
            "increment":8
          }
        ],
        "hashAlg":[
          "SHA-1",
          "SHA2-224",
          "SHA2-256",
          "SHA2-384",
          "SHA2-512",
          "SHA3-224",
          "SHA3-256",
          "SHA3-384",
          "SHA3-512"
        ]
      }
    ]
  }
]
```

Error"

```
[ {
  "acvVersion" : "1.0"
}, {
  "vsId" : 4595,
  "status" : "error during vector generation",
  "message" : "failed to generate tests for: 4595"
} 
```

celic:
The minimum keyLen is 112. I will look at the other range suggestions. 

I'm slightly concerned about runtimes for allowing an iteration count of 10mil. I'll do some tests and see what is appropriate. The standard _recommends_ using 1000 for implementations, but does not force anything really in terms of an upper or lower bound. I would discourage implementations from using an iteration count of 1, but I suppose that is left up to the original developer. 

SP800-132 specifies that the random portion of a salt must be at a minimum 128 bits. So the lowest salt allowed will be 128 bits. 

You are correct about the `"hashAlg"` here as they are used as HMAC. This could just mean updating the description in the table, or updating the property to `"macAlg"`. I'll check how other algorithms utilize MACs.

smuellerDD:
Am Mittwoch, 12. Juni 2019, 16:32:17 CEST schrieb Chris Celi:

Hi Chris,

> The minimum keyLen is 112. I will look at the other range suggestions.

Just a note, I took all upper/lower values from error messages the ACVP server 
gave me.
> 
> I'm slightly concerned about runtimes for allowing an iteration count of
> 10mil. I'll do some tests and see what is appropriate. The standard
> _recommends_ using 1000 for implementations, but does not force anything
> really in terms of an upper or lower bound. I would discourage
> implementations from using an iteration count of 1, but I suppose that is
> left up to the original developer.

Well, I do not want to get into the discussion about how to use PBKDF. But 
please note that PBKDF shall make it hard for attackers to perform a brute 
force. Thus, the number of iterations must be chosen such that it takes 
"significant" amount of time to crack it.

Commonly developers choose an iteration count that requires an execution time 
of about 100ms on a given platform.

I have implemented that in my own code (see [1] and [2]). I also have seen 
that in numerous other implementations. Considering the raw power of our 
current CPUs, the iteration count easily goes into the millions.
> 
> SP800-132 specifies that the random portion of a salt must be at a minimum
> 128 bits. So the lowest salt allowed will be 128 bits.

Ok, no problem.

> 
> You are correct about the `"hashAlg"` here as they are used as HMAC. This
> could just mean updating the description in the table, or updating the
> property to `"macAlg"`. I'll check how other algorithms utilize MACs.

[1] https://github.com/smuellerDD/libkcapi/blob/master/lib/kcapi.h#L2571

[2] https://github.com/smuellerDD/libkcapi/blob/master/lib/kcapi-kdf.c#L426

Ciao
Stephan

smuellerDD:
Using keyLen >= 128 worked and gave me the test vector.

celic:
Please let me know how your testing goes. We did not have a previous implementation of PBKDF so our crypto is new. If you have any issues validating with a "trusted" implementation please let me know.

smuellerDD:
SPEC issues:

- keyLength instead of keyLen is used in test vector

smuellerDD:
Test vector issue:

- the password field does not contain ASCII-Hex data:

```
"password" : "xjXoQcoeUZGaKVOlWLzCYmapqebAcAfeYWQunexHRJljiZJAsk",
```

smuellerDD:
Am Mittwoch, 12. Juni 2019, 17:08:57 CEST schrieb Chris Celi:

Hi Chris,

> Please let me know how your testing goes. We did not have a previous
> implementation of PBKDF so our crypto is new. If you have any issues
> validating with a "trusted" implementation please let me know.

Sure, I will first test it against libkcapi and then with OpenSSL, libgcrypt, 
GnuTLS, and PKCS11.

Ciao
Stephan

celic:
What is the vsId for the password error?

smuellerDD:
Am Mittwoch, 12. Juni 2019, 17:28:17 CEST schrieb Chris Celi:

Hi Chris,

> What is the vsId for the password error?

5168

Ciao
Stephan

celic:
The passwords are ASCII, not hex. This is stated in the spec. (`string` vs `hex`)

smuellerDD:
Am Mittwoch, 12. Juni 2019, 17:33:38 CEST schrieb Chris Celi:

Hi Chris,

> The passwords are ASCII, not hex. This is stated in the spec. (`string` vs
> `hex`)

After the interpretation of the password as a string, I was able to process 
the vector. After uploading it to the ACVP server, it indicates that all test 
responses pass.

Ciao
Stephan

celic:
That's good news. Thank you for testing.

smuellerDD:
Are you using HMAC names in hmacAlg?

Thanks for thinking about the large iterations counts. I bring that to the table to let you know. All my IUTs can operate with arbitrary count numbers. But  I can see that some IUTs set a hard lower boundary that is yet north of one million.

celic:
HMACAlg is just a name change. Not content. Also I updated the bounds as mentioned. I believe all the examples were updated. This will not take place on the demo server for a bit. I'd like to do some work on the runtime issue for larger iteration counts.

smuellerDD:
The PBKDF implementations of OpenSSL, libkcapi, libgcrypt, GnuTLS and NSS generate the data expected by the ACVP server

smuellerDD:
I added the PBKDF definitions now to other cipher definitions so that one test session contains SHA, HMAC, KDF and PBKDF.

The strange thing now is that I get the PBKDF vectors, but the majority of all other vector requests (SHA, HMAC) which normally work fail to generate a vector with the error:

```
[ {
  "acvVersion" : "1.0"
}, {
  "vsId" : 6119,
  "status" : "error during vector generation",
  "message" : "failed to generate tests for: 6119"
} ]
```

celic:
@smuellerDD I'm running some speed tests on PBKDF and am not getting promising results for large iteration counts. How quickly can one of your implementations do a 10mil iteration count on a SHA-256? (keySize = 256).

smuellerDD:
Am Mittwoch, 19. Juni 2019, 15:21:41 CEST schrieb Chris Celi:

Hi Chris,

> @smuellerDD I'm running some speed tests on PBKDF and am not getting
> promising results for large iteration counts. How quickly can one of your
> implementations do a 10mil iteration count on a SHA-256? (keySize = 256).

When testing my implementation from [1] using the test vector from RFC 6070 
with some 16 mil iterations using HMAC SHA-1, I get on my Intel i7 Broadwell:

$ time bin/.libs/kcapi -x 8 -c "hmac(sha1)" -k 73616c74 -p 70617373776f7264 -d 
16777216 -b 20
eefe3d61cd4da4e4e9945b3d6ba2158c2634e984

real    0m33,930s
user    0m11,135s
sys     0m22,630s

Note, that is a very slow implementation considering that there are 16 mil 
context switches between kernel and user. I.e. using a library, I expect that 
we are in the vicinity of 5 seconds.

[1] https://github.com/smuellerDD/libkcapi/blob/master/lib/kcapi-kdf.c#L532



Ciao
Stephan

celic:
Yeah with our current crypto, it is not feasible to do 10mil. I have added some new interfaces to perform the 10mil within a couple minutes but the hooks only exist for a couple of the HMAC algs we support. We need to talk a bit internally on whether or not this is alright (different iterationCount limits for different HMAC algs). I'd guess that the majority of real-world implementations use that subset of HMAC algs (SHA-1, SHA2-256, SHA2-384 and SHA2-512) because of their speed so that might be adequate for testing. Do you have any thoughts? @smuellerDD

smuellerDD:
Am Mittwoch, 19. Juni 2019, 18:53:53 CEST schrieb Chris Celi:

Hi Chris,

> Yeah with our current crypto, it is not feasible to do 10mil. I have added
> some new interfaces to perform the 10mil within a couple minutes but the
> hooks only exist for a couple of the HMAC algs we support. We need to talk
> a bit internally on whether or not this is alright (different
> iterationCount limits for different HMAC algs). I'd guess that the majority
> of real-world implementations use that subset of HMAC algs (SHA-1,
> SHA2-256, SHA2-384 and SHA2-512) because of their speed so that might be
> adequate for testing. Do you have any thoughts? @smuellerDD

I think it is safe to say that today's implementations use SHA1 and SHA2 (with 
SHA1 becoming sunset more and more). Also, commonly folks would use SHA256 or 
512. I have not yet seen a case with 384 in active use.

Though, most implementations have a PBKDF "block chaining implementation" that 
can work with arbitrary HMACs. And people usually try to certify all 
combinations that are technically possible.

That said, even if a SHA3 is not currently in use, it is technically possible 
and supported. And people would want to test it.

That said, what about the following: maybe have some pre-generated, but for 
the majority, let the backend generate those on the fly. Implement a counter 
for the number of PBKDFs that have an iteration count of, say, above 1mil so 
that you have some stats. When the counter reaches some numbers, you can 
revisit the approach of having these larger iteration counts pre-computed.

The key is not to disallow the larger iteration counts.

Ciao
Stephan

celic:
We would probably do a rollout of the low-hanging fruit (the mentioned HMAC algs) and work more on trying to incorporate the others. As our crypto is implemented in C#, there is only so much speed-up to gain. We could look at a native C based SHA implementation to use within our C#. This will be our likely path. For now we will support 10mil iteration count on HMACs SHA-1, SHA2-256, SHA2-384 and SHA2-512 with the intention of rolling out the others soon. This should come out on the next update to the demo environment.

Kritner:
Some updates to the registration (capabilities object) have made it out to demo.  @celic not sure if you were keeping this around for that specific change, or just for potential additional changes.

celic:
I'm not sure I got that change into the specification yet. This should remain open until I can finish that. Long story short, there is a composite 'capabilities' array containing the normal registration info similar to other algorithms.

smuellerDD:
Just for documentation, the following is a valid request:

```
     {
        "revision":"1.0",
        "algorithm":"kdf-components",
        "mode":"pbkdf",
        "prereqVals":[
          {
            "algorithm":"SHA",
            "valValue":"same"
          }
        ],
        "capabilities":[
          {
            "iterationCount":[
              {
                "min":10,
                "max":1000,
                "increment":1
              }
            ],
            "keyLen":[
              {
                "min":128,
                "max":4096,
                "increment":8
              }
            ],
            "passwordLen":[
              {
                "min":8,
                "max":128,
                "increment":1
              }
            ],
            "saltLen":[
              {
                "min":128,
                "max":4096,
                "increment":8
              }
            ],
            "hashAlg":[
              "SHA-1",
              "SHA2-224",
              "SHA2-256",
              "SHA2-384",
              "SHA2-512",
              "SHA3-224",
              "SHA3-256",
              "SHA3-384",
              "SHA3-512"
            ]
          }
        ]
      }
```

celic:
Yes exactly. Thank you :)

PR #665, opened 2019-06-18
Typo fixes

agl:
Just a couple of minor changes fixing typos that I noticed.


agl:
Each commit now updates the artifacts too. I've added an initial commit which _only_ updates artifacts, thus all the noise from changing the date etc is concentrated there for review.

celic:
Thanks for the fix on these typos.

PR #668, opened 2019-06-19
Update for vectorSetUrls

agl:
Here's the response to POSTing to /testSessions on the demo server at
the moment:

```
[ {
  "acvVersion" : "1.0"
}, {
  "url" : "/acvp/v1/testSessions/1691",
  "vectorSetUrls" : [ "/acvp/v1/testSessions/1691/vectorSets/6953" ],
  "accessToken" : "<omitted>"
} ]
```

Note that the response contains `vectorSetUrls`, not `vectorSetsUrl` as
currently documented. This change aligns the documentation with the demo
server.


agl:
Sorry, `vectorSetUrls` only appears in the response to creating a new vector set. GETting the set indeed has `vectorSetsUrl`. Have updated the PR accordingly.

PR #669, opened 2019-06-20
WIP: SP800-38G Revision 1 (draft)

Kritner:
- AES-FF1 and AES-FF3-1.
- Methods for Format Preserving Encryption https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-38Gr1-draft.pdf

Plan on using this PR as an area for comments.  The merge of this PR indicates the functionality is now present on (at a minimum) the demo server.

PR #676, opened 2019-07-10
ACVP-AES-GMAC

Kritner:
GMAC will be separated from an "implied" registration from within GCM and placed into its own algorithm.
This change will (once deployed) no longer allow zero length payloads for GCM, as that is in effect a test for GMAC.

I still need to update the symmetric spec to have this payload length change, however I will not generate the artifacts on that document (for the moment) due to there being a separate PR #669 which will conflict.


Kritner:
This change has been deployed to demo.  Note the production push of this change is not yet completed, currently planned for sometime next week

Issue #683, opened 2019-07-30
Remove obsolete large submission flows from ACVP spec 


atvassilev:
With the latest JBOS server update it is no longer necessary to handle large submissions as currently described in the spec. All definitions and references to this obsolete behavior have to be cleaned up: Section 9, Section 11.4 and Section 12.

smuellerDD:
Am Dienstag, 30. Juli 2019, 15:31:38 CEST schrieb Apostol Vassilev:

Hi Apostol,

> With the latest JBOS server update it is no longer necessary to handle large
> submissions as currently described in the spec. All definitions and
> references to this obsolete behavior have to be cleaned up: Section 9,
> Section 11.4 and Section 12.

With the latest test I conducted mid last week for Jason I still used the 
large endpoint, because the server told me so.

Ciao
Stephan

atvassilev:
Hmm, either the update has not made it to the public or something else. Let's see what the team says and will go from there.

celic:
This is still in the implementation as of today but we are working on removing the endpoint altogether. We should prepare a branch of the protocol specification with this endpoint removed though while the server is made to match.

smuellerDD:
Am Dienstag, 30. Juli 2019, 18:08:16 CEST schrieb Chris Celi:

Hi Chris,

> This is still in the implementation as of today but we are working on
> removing the endpoint altogether. We should prepare a branch of the
> protocol specification with this endpoint removed though while the server
> is made to match.

Let me know if you want me to test the regular endpoint with large data 
packets.

Ciao
Stephan

celic:
Unfortunately this may be a bit down the road as we have some other features in the works. I'll keep everyone posted via the normal calls.

Kritner:
Bumping this because we no longer require the use of the large endpoint.  

![image](https://user-images.githubusercontent.com/8814983/91051774-913ea380-e5ee-11ea-8583-1d3d8f1e679a.png)

The specification seems to already cover the fact that alternate handling *MAY* be needed, so I believe we should be safe to keep the current wording in there.  We might not be using that alternative handling, but that doesn't mean someone else implementing the protocol wouldn't.

Perhaps we can revisit at the next major revision of the spec?  With the current server implementation of the specification there'd not be an instance of a client receiving a change to their submission flow:

![image](https://user-images.githubusercontent.com/8814983/91052086-f7c3c180-e5ee-11ea-84ac-651dce04d6ab.png)

[...]

PR #686, opened 2019-08-07
 IETF review comments and remove large submission flows 

bfussell:
#683


celic:
I guess this raises a question... The server implementation has a little work to remove the large submission workflow. It will naturally lag behind the specification. I'd imagine this would be confusing for new clients/users. This seems like a fitting place for an upped version. This could be version 01. That way we can declare the server implementation is compliant to version 00, and soon will move to version 01. We can discuss tomorrow if needed.

bfussell:
Sure, we can discuss tomorrow.

PR #727, opened 2019-10-28
KAS/KTS-IFC Specification SP800-56Br2

### LINKED ISSUE #726: Introduce Specification for SP800-56Br2
SP800-56Br2 (KAS/KTS-IFC) is in the works, a specification is needed for this functionality.

* This update will also make use of the SP800-56Cr1 specification
  * to be applied to (a yet to be created) SP80056Ar3 testing at a later date.

Kritner:
~Work in progress DO NOT MERGE~

[SP800-56Br2](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Br2.pdf)
[SP800-56Cr1](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Cr1.pdf)

* Specification for upcoming ACVP functionality as it pertains to SP800-56Br2 (KAS/KTS-IFC)
* Also implements functionality from SP800-56Cr1 as it pertains to SP800-56Br2
* Use PR as a place for comments
* Merge of PR indicates the functionality is present on (at a minimum) the demo server
* closes #726


Kritner:
I *think* that's it for the first pass through of this specification.  

One odd thing to note about this algorithm testing in comparison to others is in order to accommodate our testing workflow (receive registration, provide prompt file, receive responses, provide disposition) we need to take in a set of public keys from the IUT *at the time of registration*.  There should be at LEAST one public key for every combination of modulo and private key format supported.

The IUT provided public keys will be used when constructing the AFT tests, and the ACVP server will be generating keys to use on behalf of the ACVP server AND IUT for the VAL tests.  The public keys are used for encrypting `Z` values into a `C`; when providing the `C` value to the IUT in the prompt file, the public key that was used to encrypt the `Z` value will also be included.

There is still some time before this will be ready to make it out to demo, but interested parties should try to make a point to review and provide feedback.

Thanks!

Kritner:
KAS/KTS-IFC is now available for testing on the demo server.

PR #730, opened 2019-10-29
Adds requiresEmptyIv property for Feedback KDF

### LINKED ISSUE #729: Implementations may not support IV for Feedback KDF
Some implementations do not support an IV for Feedback KDF (SP800-108). Under ACVP, the current `supportsEmptyIv = true` flag will generate groups that have an empty IV and groups that utilize the IV.

I propose a new boolean `requiresEmptyIv` to go alongside the existing boolean that will behave identically but remove the groups that utilize an IV. I like an additional flag to keep existing clients compatibility without work. This flag would be completely optional and assumed to be `false` if not provided.

celic:
`requiresEmptyIv` sits next to `supportsEmptyIv`. The difference being, if `requiresEmptyIv` is set to true, no groups that utilize an IV for Feedback KDF will be included in the vector set. If `supportsEmptyIv` is set to true while `requiresEmptyIv` is set to false, then both groups that utilize IV and do not utilize IV for Feedback KDF will be included.

Fixes #729

### DIFF
--- a/artifacts/acvp_sub_kdf108.html
+++ b/artifacts/acvp_sub_kdf108.html
@@ -396,7 +396,7 @@
 <link href="#rfc.authors" rel="Chapter">
 
 
-  <meta name="generator" content="xml2rfc version 2.21.1 - https://tools.ietf.org/tools/xml2rfc" />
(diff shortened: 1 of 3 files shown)


celic:
Fixes #729

celic:
This is out on demo. Please let me know if there are issues.

PR #824, opened 2020-04-06
KAS spec updates

### LINKED ISSUE #814: KAS_IFC documentation comments
After implementing SP800-56B rev2 KAS/KTS IFC, I have the following comments:

3.4.1.1.1: Modulo -> modulo

3.4.1.1.1: FixedPublicExponent -> fixedPubExp to be consistent with RSA

3.2: "algorithm": the specification seems to be internally inconsistent here. It allows the specification of only one algorithm, but the "scheme" definition allows a list of objects which may include KAS and KTS. I think there needs to be one and only valid value for algorithm.

3.4.1: names of schema inconsistent with examples (see capitalization)

3.4.1.1.1: the modulo size of 7168 seems to be not accepted:

```
{
  "acvVersion" : "1.0",
  "error" : "Property value(7168) for (scheme/KAS1-Party_V-confirmation/keyGenerationMethods/rsakpg1-basic/modulo) in Algorithm (KAS-IFC (null)) is not valid. Validation: (Values must be one of: 2048, 3072, 4096, 6144, 8192)"
}
```

Kritner:
- Updates fixedInfoPatternConstruction sample to contain even number of literals
- closes #814

### DIFF
--- a/artifacts/acvp_sub_kas_ecc.html
+++ b/artifacts/acvp_sub_kas_ecc.html
@@ -1511,7 +1511,7 @@
 <p></p>
 
 <ul>
-<li>literal[123456789ABCDEF] <ul><li>uses the specified hex within "[]". literal[123456789ABCDEF] substitutes "123456789ABCDEF" in place of the field</li></ul>
+<li>literal[0123456789ABCDEF] <ul><li>uses the specified hex within "[]". literal[0123456789ABCDEF] substitutes "0123456789ABCDEF" in place of the field</li></ul>
 <p> </p>
 </li>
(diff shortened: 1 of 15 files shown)


Kritner:
@smuellerDD I think this addresses everything in #814 except the comment from https://github.com/usnistgov/ACVP/issues/814#issuecomment-609967129

What are you thoughts on that?

Kritner:
Oh actually I guess it did not address: 
> 3.4.1.1.1: the modulo size of 7168 seems to be not accepted

either... let me get back to you on that.

smuellerDD:
Applied changes look good.

PR #961, opened 2020-08-19
Fix column count in symmetric cipher table 6

RKinsey:
Nine columns were declared in the Miscellaneous Block Cipher Algorithm Capabilities Applicability Grid (Table 6 in the block cipher document), but ten headers are used. The table doesn't display correctly with this mismatch, so I added the last column to the declaration.

### DIFF
--- a/src/symmetric/sections/05-capabilities.adoc
+++ b/src/symmetric/sections/05-capabilities.adoc
@@ -117,7 +117,7 @@ The following grid outlines which properties are *REQUIRED*, as well as the poss
 
 The following grid outlines which properties are *REQUIRED*, as well as the possible values a server *MAY* support for each miscellaneous block cipher algorithm:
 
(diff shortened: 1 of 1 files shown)


src/symmetric/sections/05-capabilities.adoc
```diff
@@ -117,7 +117,7 @@ The following grid outlines which properties are *REQUIRED*, as well as the poss
 
 The following grid outlines which properties are *REQUIRED*, as well as the possible values a server *MAY* support for each miscellaneous block cipher algorithm:
 
-[cols="<,<,<,<,<,<,<,<,<"]
+[cols="<,<,<,<,<,<,<,<,<,<"]
```
Kritner, line 120: Isn't the cols annotation mostly superfluous anyway?  I had mostly been removing it when I saw it when going through the initial conversion.

Kritner:
thanks!

PR #963, opened 2020-08-20
Fixed example symmetric vector display bug

RKinsey:
Corrected an errant linebreak in  `src/symmetric/sections/97-examples.adoc` that swapped the vector and non-vector formatting in all text after the AES-GCM responses.


RKinsey:
Also fixed a table in `src/mac/sections/06-hmac-test-vectors.adoc` with an extra column in its declaration

PR #1004, opened 2020-09-17
Updates to cshake and derivatives MCT psuedocode

Kritner:
- #904


dghgit:
Hi @Kritner, this looks okay. I think calling out the assumed zero padding was a good idea. Let me know when it's in test and I'll give it a whirl, the required changes on my end are pretty trivial. Thanks.

PR #1010, opened 2020-09-21
States ivLen will always be 96 bits in XPN groups

### LINKED ISSUE #1009: AES-XPN: Says "Within domain" for ivLen when should say "96"?
Table 11 in the block cipher spec indicates that the iv length for tests from a server for AES-XPN must be "within domain". My understanding is that XPN has a static 96 bit iv length. Could this be clarified?

Thanks,
Andrew

celic:
Fixes #1009


celic:
Closes #1009

PR #1016, opened 2020-09-22
Large Data Tests for SHA1,2,3

celic:
Large Data Tests are optional tests for all hash functions where an IUT performs a hash on multiple gigabytes of data. These specification updates provide all the needed information and examples. 

As our JSON prompts cannot handle sending a 4GB file to clients, we use a new type of JSON object to communicate a large message. An example is below. In this case, the "content" is repeated as many times as is required until the full length is achieved. Then the content would be passed to the hash function to produce a digest. 

```
"largeMsg": {
    "content": "DE26",
    "contentLength": 16,
    "fullLength": 42949672960,
    "expansionTechnique": "repeating"
}
```

Discussion of these tests and their motivation is available here https://eprint.iacr.org/2019/1421.pdf 

These are undergoing internal testing at the moment but should be included in an upcoming demo release cycle in October.


smuellerDD:
Could you please either add some words in the specification or explain somewhere else when the LDT test type is mandated, if at all?

I personally think that not many users would want to support linear buffers > 4gb.

Thus, what happens if the API of the IUT technically would be able to support linear buffers >4gb, but the environment may not allow it (e.g. the RAM of the test system is 4GB or less)? Or the OS does not allow that much memory to be allocated to a single process, etc...

celic:
LDT is an optional test. The ability of the module to handle large inputs demonstrates robustness on an algorithm certificate. As tests are intended to happen on the environment, the capability registration MUST reflect that. So even though an implementation may appear to handle large inputs just fine, if the environment does not allow it, that is what is being validated. The implementation on a specific operating environment. 

There are cases where users would not want to support large linear buffers yes. They must declare that in the capabilities and it will show up on the algorithm certificate. 

What we are looking to prevent, is an IUT claiming they support ALL data sizes because they met the largest test vector provided by ACVP. These two ideas are not equivalent and can lead to danger in real-world use. There are times, in particular for hashes, where the message is provided by something external to the crypto module itself (such as signature verification). If a large data test is not performed, I would expect some bounds check to exist to prevent accepting such a message for signature verification.

celic:
We are interested in looking at particular boundaries for these tests. Right now in our testing, the group generates a single test at 5GB. (Intentionally more than 4GB as this is a boundary for 32-bit vs 64-bit architectures). We may expand this to allow a client to specify which GB boundaries they would like to test from 2-8GB.

celic:
Latest change modifies the registration from a boolean to an integer array. An example is the following... 

```
{
    "algorithm": "SHA2-256",
    "revision": "1.0",
    "messageLength": [{"min": 0, "max": 65535, "increment": 1}],
    "performLargeDataTest": [1, 2]
}
```

where the above registration would receive two Large Data Tests. One with a 1GB message and one with a 2GB message.

smuellerDD:
Am Montag, 28. September 2020, 22:31:52 CEST schrieb Chris Celi:

Hi Chris,

> Latest change modifies the registration from a boolean to an integer array.
> An example is the following...
> 
> ```
> {
>     "algorithm": "SHA2-256",
>     "revision": "1.0",
>     "messageLength": [{"min": 0, "max": 65535, "increment": 1}],
>     "performLargeDataTest": [1, 2]
> }
> ```
> 
> where the above registration would receive two Large Data Tests. One with a
> 1GB message and one with a 2GB message.

May I ask for slight update to be consistent with the rest: It is not clear to 
me whether we have GBit or GByte. In addition, are we talking about GiB or GB? 
I.e. 2^30 or 10^9? Based on these questions, why not use a domain here with a 
min of 2^30 and a max of your choice to allow also, say, 3.5 GB by using a 
JSON 64-Bit integer.

Thanks a lot.

Ciao
Stephan

celic:
This uses 2^30 bytes as 1GiB, so 2^33 bits as 1GiB. I can clarify this in the spec though because I called it a GB not a GiB. The intention is to test the boundary between a 32-bit and 64-bit system, so GiB is the correct term. 

For the domain, we didn't feel it was informative to look at a validation listing with such a large integer. It is difficult to tell how large 68,719,476,736 bits is in a domain.

celic:
These tests are on demo. They are optional, not specifying the array will lead to the tests not being included.

Kritner:
This change is on demo in release [v1.1.0.13](https://github.com/usnistgov/ACVP-Server/releases/tag/v1.1.0.13).

Kritner:
This change is now on prod in release https://github.com/usnistgov/ACVP-Server/releases/tag/v1.1.0.13

Issue #1095, opened 2020-12-11
Introduce specification for SP800-56Cr2


Kritner:
Introduce specification outlining testing against [SP800-56Cr2](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Cr2.pdf).

The testing should be quite similar to current SP800-56Cr1 testing, with a few additions:

* Introduction of new fixed info pattern `t`, which represents a secondary shared secret between party u and party v.
  * the implementation details are not specified within the SP800-56Cr2 document, but the feature can be "tested" by injecting the `t` value provided in the prompt file (when registered for using a fixed info pattern option of `t`.
* Introduction of Multi Expansion KDF testing for TwoStep KDF (including HKDF)
  * Section 5.3 "Randomness Extraction followed by Multiple Key Expansions" within SP800-56Cr2 describes running the "extract" portion of the KDF a single time, then multiple iterations of the "expand" step.  Testing will be implemented so that IUTs ***MAY*** register to test against multi expansion, but since the details of what goes into some parameters of multi expansion (namely fixed info construction throughout each iteration), the IUT ***MUST*** (if the IUT supports multi expansion) be able to inject the per iteration parameters generated by the ACVP server.

Kritner:
This change is on demo in [v1.1.0.15](https://github.com/usnistgov/ACVP-Server/releases/tag/v1.1.0.15)

PR #1096, opened 2020-12-11
Testing specification for SP800-56Cr2

Kritner:
* new testing for HKDF, OneStepKdf, and TwoStepKdf
* Renamed SP800-56Cr1 documents to "SP800-56C", and referenced Cr1 and Cr2 within the document when applicable.
  * The testing between Cr1 and Cr2 is *mostly* the same, so rather than create brand new documents, the "C" document now supports both.
* https://github.com/usnistgov/ACVP/issues/1095


smuellerDD:
Shouldn't the saltLen parameter be added to the TwoStep and
OneStep/AuxFunction defintiion to be consistent with the rest as they require
specifying the salt type?

Kritner:
I suppose if it becomes a problem they could be added, though for the moment Cr1 does not have this capability either.  The server provides the random value or the "default" value as a property within the prompt file, and is the length of the output of the function being utilized (or 128/256 for kmac)

Issue #1049, opened 2020-10-28
Support for a specific use case such as RFC3686


ykalensc:
How is NIST planning to address implementations that only support a specific use case such as RFC3686?

celic:
What in the current implementation would cause an issue here?

This is more a question for the CAVP rather than this project, but in the past we have added "conformances" to an algorithm test when a client wishes to test an ACVP algorithm and adhere to an additional standard on top of the NIST standard for that algorithm. See ECMA conformance for AES-CCM. This allows the server to build specific test cases with properties it might not otherwise allow, with the trade-off that a client must specifically request such a test.

ykalensc:
Current crypto module that I'm working with only has an implementation of rfc3686(ctr(aes)), however, ACVP test vector is only available for ctr(aes). I was wondering how to get a test vector for specific cases such as rfc3686(ctr(aes)) from NIST server.  

I'm not familiar with ECMA conformance.. Is there a link that you can provide?

celic:
Please send me an email so we can discuss some details here. christopher.celi (at) nist.gov

smuellerDD:
The Linux kernel rfc3868() template is not to be tested. All it provides is a special interpretation of the input data, i.e. where and how it finds the IV/CTR or the key. Thus, you can safely ignore this template and only test ctr(aes).

ykalensc:
Hi smueller, 
we are using a 3rd party vendor module and unfortunately, they only supports rfc3686 compliant aes-ctr. That's why I was asking for test vector for specific case.

smuellerDD:
Then what about formatting the provided data as needed by the template? IIRC the key and/or counter needs to be split or merged to feed it to the template. For example, that is also to be done when testing the rfc4106 template.

Ciao
Stephan

—
Cui bono?

ykalensc:
If I understand RFC3686 correctly, there are 2 parts that would need modification.
1. RFC3686 counter block is formed with Nonce(32bits) + IV(64bits) + ONE(32bits). 
However, CTR(AES) test vector doesn't guarantee the counter blocks with ONE at the end of 32 bits, and I don't think we can format this portion (this will change the expected result as well).
2. RFC3686 expects Nonce(32bits) + Key.. and we can format the keying material combined with nonce.

Because of the option 1, formatting the ctr(aes) test vector doesn't seem to be an option for us. 
[...]

PR #1101, opened 2020-12-22
Specification updates for RFC3686

Kritner:
* https://github.com/usnistgov/ACVP/issues/1049


smuellerDD:
Am Dienstag, dem 22.12.2020 um 11:15 -0800 schrieb Russ Hammett:
> * https://github.com/usnistgov/ACVP/issues/1049
> You can view, comment on, or merge this pull request online at:

Although I praise the addition, I would like to note that NIST now opened the
can of worms for other requests to cover other such "conformances".

As immediate conformances I see:

- RFC4309 - CCM

- RFC4106 - GCM (albeit that can be handled with the existing vectors when
massaging them a bit)

- RFC4543 - GCM

- CCMP

- GCMP

Thanks
Stephan

Kritner:
Yeah @smuellerDD it is potentially a can of worms.  I'm not sure if there's a better way to approach this or not... conceivably there will be instances where RFCs will be very particular in the construction of values that get plugged into the crypto, that our normal testing would not be able to accomodate.  This is one of those instances, but there could very likely be more.  We can try to address those when they occur, but also should consider if there's a better way to handle situations like this.

PR #1125, opened 2021-02-08
Fixed invalid revision string in KAS-ECC-SSC

RKinsey:
Original string was listed as "SP800-56Ar3" instead of "Sp800-56Ar3". The server is case-sensitive, so copying the revision field creates a somewhat confusing error if you aren't already familiar with the correct capitalization.


Kritner (changes requested):
Looks like there's some instances of this in the `03-supported.adoc`, `06-test-vectors.adoc`, and `07-responses.adocs` as well.

LMK if you want to submit those changes (or not and I'll take care of it) prior to merge.

Thanks!

this probably also applies to the non-SSC kas, though I have not confirmed that.  No need to add that to the scope of this PR, unless you feel like it ;)  One of us will get to it eventually if not.

RKinsey:
Oops, should've done a search. I'll do it, won't take very long.   
Do you also want the `spec-algorithm`, `abbr`, and `algo-short-name` fields changed?

RKinsey:
Pushed all the changes other than `spec-algorithm`, `abbr`, and `algo-short-name`

Kritner (approved):
Thanks!

Issue #1179, opened 2021-05-19
Key wrap test of direction and cipher combinations


swkeypair:
[SP 800-38F defines KW-AE as "encrypt" using the forward cipher](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-38F.pdf#page=22), and defines unwrapping KW-AD as "decrypt" using the inverse cipher.  However, in ATF request vectors seemingly inappropriate combinations of "direction" and "kwCipher" are tested:

> "testType": "AFT",
> "direction": "encrypt",
> "kwCipher": "inverse",

Is this the correct interpretation of the standard?

For example, why is the encrypt operation to be tested using the inverse cipher when it is not defined as such in KW-AE?

Kritner:
From https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-38F.pdf#page=16

![image](https://user-images.githubusercontent.com/8814983/119363431-cc542d00-bc7b-11eb-8b75-79fa3e0fd15c.png)

The underlying cipher of the key wrapping algorithm can be "cipher" or "inverse", then there is the `W` (wrap / auth encrypt) and  `W1` (unwrap / auth decrypt).

This gives several "combinations" of parameters that go into what the IUT performs as it relates to the test group / test case

| Operation | Cipher | underlying AES operation |
| --- | --- | --- |
| Wrap (direction encrypt) | Cipher | aes encrypt |
| Wrap (direction encrypt) | Inverse | aes decrypt |
| Unwrap (direction decrypt) | Cipher | aes decrypt |
| Unwrap (direction decrypt) | Cipher | aes encrypt |

![image](https://user-images.githubusercontent.com/8814983/119365706-35d53b00-bc7e-11eb-99bd-73e629899daa.png)

In the above `CIPH_k` can be either "cipher" (encrypt) or "inverse" (decrypt). 

Hopefully that makes sense... basically you can "Wrap" a key using a aes decrypt as the cipher (this would be the "inverse" cipher invoke), then the "Unwrap" would utilize the opposite of that - encrypt.  Vice versa also works using the encrypt operations as the "wrapping" cipher, and decrypt operation as the "unwrapping"; I called it out in the opposite direction first, because it seems less intuitive to utilize the algorithm in this operating mode (IMO).

swkeypair:
OK - please confirm or correct:
[...]

PR #1223, opened 2021-08-17
dmchodnicki symmetric aes kw

dmchodnicki:
Adding Wrap/Unwrap Operations table for additional clarity.  Resolves ACVP #1179.


Kritner (changes requested):
I'm not sure what exactly happened here, but between #1222 and #1225, it looks like everything was created off of the same "already worked on" branch, rather than starting from master as its root.  This has caused the "most recently submit" PR to have the changes from each individual PR.

This is not *necessarily* a problem, but it's best to start branches from master, rather than starting branches for "new features" on branches from "new features" that have not yet been merged.

I don't know if it's worth redoing the PRs, but I could help you do so if you'd like, it would make the history look a bit cleaner, and you'd be able to better attribute PRs to specific issues.

You can tag the issues with "#" then the issue number, like "#1223" would tag this PR as an example.

Kritner (approved):

Issue #92, opened 2017-03-24
Specify Field order?


hbooth:
Although JSON does not require it, as I am going through implementation I am wondering if we shouldn't require a particular field order. I am trying to be somewhat accepting and will ignore any unknown fields, but there are a few fields that I really need to know in a particular order. Largely, the order they are described in the specs are the order that I need. For the record, I am currently trying to process the input as a stream, if we cannot (or do not) want to specify the order then I will need to use a different implementation (not hard) but the consequence for us will be one of memory usage. I will need to load the whole file in memory and perform some random access. (The library I am using has the ability to create a tree-node representation and I can switch to that.)

A couple of examples:
 - the acv_version field must be first or I don't know how to process the rest of the file.
 - need the algorithm field before any of the other fields and so on

I do need to know sooner rather than later though how we want to address this moving forward.

atvassilev:
I sympathize with the issue you are facing but am leery about enforcing order since JSON does not require it. My gut instinct tells me enforcing an order increases the brittleness of the protocol and this is not a good thing. 

Concerning the memory resource utilization, I guess I don't understand the memory resource problem. My guess is that you will be expanding to hold the entire file in memory anyway at some point, even if you had the fields in the right order and the file is processed as a stream. So, if I my guess is correct, the question is do we expand upfront and then process, or first process a little and expand in the end. 

If the library you are using allows sharing the tree-node representation around all components of the public server that need it, then the memory issue washes out and this becomes a matter of processing flow management. The main visible difference between the two processing alternatives is that you will be able to respond quicker to errors in the version in one case than the other. But this is a small consolation given the problems associated with enforcing an order.

hbooth:
Now we are getting a little more into the weeds, but I am reading in the input and then converting to a "neutral" representation. During this processing I perform all validation and throw away anything I don't need to minimize my internal representation. At the moment, I can likely work around everything but acv_version and keep the stream processing. I do fundamentally need to know what the version of the protocol I should use to process the input. Most other protocol/file formats I know of sticks the version information at the beginning of the format. I could theoretically, read everything into the tree-node representation, validate the tree nodes, and then wrap these nodes with wrapper classes to keep the memory footprint small, but that would complicate a number of things. An alternative would be to convert to tree node, process and validate building the neutral representation with the result of two in-memory representations temporarily. I could also add some caching to process the file until I find acv_version in the stream, but then worst-case a client could stick the acv_version at the end and that wouldn't buy me much.

[...]

PR #1231, opened 2021-08-26
XOF Spec ParallelHash updates.  Resolves ACVP-Server #92.

dmchodnicki:
Closes https://github.com/usnistgov/ACVP-Server/issues/92


Kritner:
Note that this is https://github.com/usnistgov/ACVP-Server/issues/92, not #92.  The full URL to the issue needs to be referenced in a commit/PR when the issue is from a separate repository

PR #1285, opened 2022-01-20
Update SHA-MCT pseudocode

schneidersoft:
I found this code extremely difficult to read.
The Indentation was inconsistent.
Depending on the interpretation of the for loop semantics [0, 99) vs [0, 99],  the inner loop discards the MD[1003] product or the outer loop only produces 99 iterations.


livebe01:
@schneidersoft thanks for the catches and edits on this!

PR #1291, opened 2022-01-25
Fix swapped naming of test case fields

cipherboy:
In the present version of the spec, table 5 names the fields of the ACVP
test cases. However, the order of the parts of the name (namely,
hello/finished and server/client) are swapped from the value given in
the example and by the Demo server. E.g., `serverHelloRandom` is in
practice called `helloServerRandom`. Fix the two to match.

`Signed-off-by: Alexander Scheel <alex.scheel@hashicorp.com>`

---

Must confess this threw one for the better part of two days as I tried to figure out why HKDF was giving wrong answers in an otherwise working TLS implementation. :-) I had copied the JSON field names from the table, rather than look closely at the test vectors themselves, meaning all the `*Random` fields were empty.


livebe01:
Eck, sorry about that!

cipherboy:
No problem, thanks for the merge! :-)

PR #1318, opened 2022-03-14
Update documents to use latest NIST Metanorma toolchain

ronaldtse:
NOTE: Background and additional information provided in metanorma/metanorma-nist#300.

### Clean up bibliographic entries

When multiple documents refer to the same hard-coded bibliographic entries, we can share the same entry.
* metanorma/ACVP@290584b70ae7faea7817b704a848630dc345bd32

Addendums (the only entry is SP 800-38A-Add)  are now supported in reference auto-fetch:
* metanorma/ACVP@f0eba7c5ce2b8062f461715534bc4dde68b6eb3e

Drafts are now supported in reference auto-fetch:
* metanorma/ACVP@bfab8dcf0cb690029145560946f5130d61727afe

Some clean up of AsciiBib anchors:
* metanorma/ACVP@4bfca36c8ffac1f978a444a09790477b92fd913a

### Migrate ACVP documents to NIST CSRC format

The full conversion to the official Metanorma NIST format:
* metanorma/ACVP@330308369a89033b6c5f4626a74bbc77ae6a21c0

There are differences from the previous workflow:
* Instead of generating output into TXT, Metanorma NIST provides the CSRC PDF format
* Output for the HTML version has the NIST approved appearance, including the official NIST header/footer.

### Clean up GitHub Pages deployment of ACVP

I have also cleaned up the deployment of `index.html` so that GitHub Pages now automatically generates the HTML and CSS automatically.

This implementation now keeps the NIST headers/customizations without needing to keep `index.html` as part of the repository.

* metanorma/ACVP@d6341de436137253cdcb201e40d17502a4edf116

Notice the NIST headers/footers don't work right now. This is because of CORS -- the NIST headers/footers script only works on https://pages.nist.gov, and when loaded on other domains (except 127.0.0.1) they will fail to execute. i.e. when this is deployed at the correct location, the headers/footers will work.

Screenshot 2022-03-14 at 8 25 24 AM

## WARNING: After merging this PR

After merging the PR, you need to do two things.

### Update Jekyll's `_config.yml`

Because of how ACVP is deployed to a custom domain name at `https://pages.nist.gov`, additional configuration may be needed at GitHub Pages after merging the PR.

You may need to add this config to your `_config.yml` file in the ACVP repository, which is used for GitHub Pages / Jekyll to generate the site (the code is currently commented out at https://github.com/metanorma/ACVP/blob/d6341de436137253cdcb201e40d17502a4edf116/_config.yml#L5-L8):

```yaml
# In _config.yml, add anywhere
url: https://pages.nist.gov
baseurl: "/ACVP" # the subpath of your site, e.g. "/blog"
repository: "usnistgov/ACVP"
```

### Enter a Personal Access Token into the "Secrets" section that has access to Metanorma NIST

Metanorma NIST is only available to and accessible by NIST staff.

A GitHub [Personal Access Token (PAT)](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) is used to validate access. An account that the PAT is generated from must already have access to the `metanorma-nist` gem and the `metanorma-nist` docker container.

NOTE: Please contact the CSRC publications team if additional access is required.

In the GitHub Actions workflow, there are two locations where the private token is used:

* https://github.com/metanorma/ACVP/blob/d6341de436137253cdcb201e40d17502a4edf116/.github/workflows/docker-build.yml#L15
* https://github.com/metanorma/ACVP/blob/d6341de436137253cdcb201e40d17502a4edf116/.github/workflows/docker-build.yml#L33 (this "step" is optional, it is only necessary if other private gems is used)

You will need to set this token on the "Settings > Secrets" page at usnistgov/ACVP, see [this guide](https://docs.github.com/en/actions/security-guides/encrypted-secrets) for details.


ronaldtse:
@livebe01 the current PR is failing due to the missing access token here: https://github.com/usnistgov/ACVP/pull/1318/files#diff-3414847e2ad632333f775cabb810f0dc0df61a570365df34750a08b00912fe82R15

Without this token the build will not work.

jbrock24:
Hi @ronaldtse, we just wanted to touch base with you regarding this PR; this is still within our radar and on our docket, appreciated!

ronaldtse:
Thanks @jbrock24 for the update! This PR probably needs a bit of update, we'll do that and ping you again.

jbrock24:
> Thanks @jbrock24 for the update! This PR probably needs a bit of update, we'll do that and ping you again.

No problem. We'll let you know, thanks again!

livebe01:
Thanks @ronaldtse, we'll get it pushed through this time.

ronaldtse:
@livebe01 @jbrock24 this PR is now rebased and works. The above mentioned steps are still necessary and require your attention. Thanks!

livebe01:
Great, thank you!

PR #1321, opened 2022-03-24
Modify the description of the requirement for the Large Data Test

locksmithone:
Modify the description for the requirement of the single call for the LDT. Regardless of the interfaces available to the operator of the IUT, if the LDT is enabled, the entirety of the large message must be processed through a single Update() or Hash() call (and not multiple update calls).


locksmithone:
Update was done outside of the pull request. Pull request not needed anymore.

PR #1332, opened 2022-04-27
AES XTS Revision Update

Kneckter:
When AES XTS 1.0 and SHA3 1.0 were deprecated, only SHA3 received a note in the documentation. This update adds the note to the AES XTS 1.0 revision so people know that only 2.0 is accepted by the server.


livebe01:
Thanks for the help on this!

PR #1362, opened 2022-09-02
[shake-mct] Fix typo

kriskwiatkowski:
I think SHAKE takes ``MSG`` as an input. Otherwise ``M`` is not defined.

### DIFF
--- a/src/sha3/sections/04-testtypes.adoc
+++ b/src/sha3/sections/04-testtypes.adoc
@@ -57,7 +57,7 @@ For j = 0 to 99
         MSG[i] = 128 leftmost bits of MD[i-1]
         if (MSG[i] < 128 bits)
             Append 0 bits on rightmost side of MSG[i] til MSG[i] is 128 bits
-        MD[i] = SHAKE(M[i], OutputLen * 8)
+        MD[i] = SHAKE(MSG[i], OutputLen * 8)
 
         if (i != 1000)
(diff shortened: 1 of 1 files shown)


celic:
I'm going to reject because we also need to remove L62 and fix the indentation. I'll make this fix though, thank you for the PR.

kriskwiatkowski:
@celic Additionally there is ``M`` in line 47. Should be ``MSG``. Also line 50 has a typo, ``til`` -> ``till``.

kriskwiatkowski:
@celic hmm... I'm not sure really. Look, with definition like this:

```
    For i = 1 to 1000
        MSG[i] = 128 leftmost bits of MD[i-1]
        if (MSG[i] < 128 bits)
            Append 0 bits on rightmost side of MSG[i] til MSG[i] is 128 bits
        MD[i] = SHAKE(MSG[i], OutputLen * 8)

        RightmostOutputBits = 16 rightmost bits of MD[i] as an integer
        OutputLen = minOutBytes + (RightmostOutputBits % Range)

    Output MD[1000], OutputLen
```
OutputLen, at the last line may not be equal to length of MD[1000]. Sure, it should be like that?

celic:
In this case the outputlen should just be the length of the output, it does
not necessarily need a variable for that stage. It could be updated to be
clearer, certainly.

PR #1365, opened 2022-09-12
Updating broken links from KDA algorithms.

locksmithone:
Changing some links from github (which do not yet exist) to point to pages.nist.gov.


livebe01:
Thanks @locksmithone. It looks like we missed those edits when we migrated the specs to pages.nist.gov. Appreciate it!

PR #1413, opened 2023-02-08
Update Install Link to Metanorma

okilker:
The current Install link is broken, produces a 404 error

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Using Metanorma is not required to view the documents in the GitHub repository. 
 
 Metanorma is a library for writing standards. It can compile `.adoc` files into multiple common standards formats including the IETF's RFC format.
 
-Use the instructions here to set up Metanorma: https://www.metanorma.com/author/topics/install/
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -42,7 +42,7 @@ Using Metanorma is not required to view the documents in the GitHub repository.
 
 Metanorma is a library for writing standards. It can compile `.adoc` files into multiple common standards formats including the IETF's RFC format.
 
-Use the instructions here to set up Metanorma: https://www.metanorma.com/author/topics/install/
+Use the instructions here to set up Metanorma: [https://www.metanorma.com/author/topics/install/](https://www.metanorma.org/install/)
```
Kritner, line 45: Looks like this was made a MD link, but only the "link" portion was updated.  Please update the "display" portion as well (or just fall back to the valid URL w/o the MD style link).

Kritner (changes requested):

README.md
livebe01: done!

PR #1414, opened 2023-02-08
Adds LMS

celic:


Kritner (changes requested):
Will probably want to add pseudocode for the deterministic keygen/signatures aspects of the algorithm.

I think there was (perhaps) a deterministic method specified in either the RFC or SP around key generation.  But (unless it was removed) there is also a deterministic method for `c` generation for the creation of signatures?

I may be misremembering some things here, but wanted to at least call it out!

livebe01:
The LMS modes are camel case in some parts of the spec and pascal case in others. Do you want them to be camel case like RSA or something else?

livebe01 (commented):
Apart from the question/comment I made about mode casing, looks good!

celic:
Will update casing to be consistent with other signature algorithms. Will add discussion on deterministic tree/signature generation.

celic:
Please double check the casing to see if I missed anything. Also added a note in the `Requirements Covered` section on the pseudorandom LMOTS key generation. It's added as a *MAY* but it is how the NIST ACVTS instance will test LMS keyGen. I did not add discussion on pseudorandom `C` because it is not relevant for LMS. It will be relevant for HSS though.

Kritner (approved):

livebe01:
> 

@celic, can you double check that the doc still builds locally for you? It built fine for me yesterday morning, but errors out now w/ the latest changes.

celic:
Everything builds just fine on GitHub Actions

celic:
Messed up an RFC reference. Fixed.

livebe01:
Looks good to merge to me.

celic:
Please merge when LMS goes to Demo.

PR #1452, opened 2023-06-30
Updates SHA1/2 MCT testing to remove dependence on registration values

livebe01:
Rewrites the SHA1/2 MCT testing to remove the need for an IUT to have knowledge of values from the registration.

### DIFF
--- a/src/sha/sections/04-testtypes.adoc
+++ b/src/sha/sections/04-testtypes.adoc
@@ -44,9 +44,9 @@ For j = 0 to 99
     For i = 0 to 999
         MSG = A || B || C
         if LEN(MSG) >= LEN(SEED):
-            MSG = TruncateToSize(MSG, LEN(SEED))                
+            MSG = leftmost LEN(SEED) bits of MSG
         else:
-            MSG = MSG || CreateZeroBitStringOfLength(LEN(SEED) - LEN(MSG))
+            MSG = MSG || LEN(SEED) - LEN(MSG) 0 bits
         MD = SHA(MSG)
         A = B
         B = C
--- a/src/sha/sections/04-testtypes.adoc
+++ b/src/sha/sections/04-testtypes.adoc
@@ -19,9 +19,7 @@ There are two types of tests for SHA-1 and SHA-2: functional tests and Monte Car
 
 The MCTs start with an initial condition (SEED which is a single message) and perform a series of chained computations. The algorithm is shown below.
 
-NOTE: The MCTs were updated, 3/2023, to support the case where !SupportedMessageLengths.Contains(3*digestSize), a limitation of the original MCT design. The change is backwards compatible. 
-
-NEW NOTE: The MCTs were updated, 7/_/2023, to support the case where !SupportedMessageLengths.Contains(3*digestSize), a limitation of the original MCT design. As part of this change, a new property, mctVersion was added to the MCT test case JSON (See ___ section of this spec). When a value of "standard" is provided in the prompt file, the standard MCT test is run. When a value of "conistent" is provided in the prompt file, the consistent MCT test is run. When the standard test is run, the MSG hashed will always be 3 * the length of the SEED. When the consistent test is run, the MSG hashed will always be the length of the SEED.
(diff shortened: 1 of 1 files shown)


src/sha/sections/04-testtypes.adoc
```diff
@@ -36,36 +36,23 @@ For j = 0 to 99
...
-                C = MD || CreateZeroBitStringOfLength(LEN(SEED) - digestSize)
-            Else:
-                C = TruncateToSize(MD, LEN(SEED))
-        Output MD
-        SEED = C
+For j = 0 to 99
+    A = B = C = SEED
+    For i = 0 to 999
+        MSG = A || B || C
+        if LEN(MSG) >= LEN(SEED):
```
celic, line 46: Don't really need `=` here

src/sha/sections/04-testtypes.adoc
```diff
@@ -36,36 +36,23 @@ For j = 0 to 99
...
+            MSG = TruncateToSize(MSG, LEN(SEED))                
+        else:
+            MSG = MSG || CreateZeroBitStringOfLength(LEN(SEED) - LEN(MSG))
```
celic, line 49: `CreateZeroBitStringOfLength` should just be `LEN(SEED) - LEN(MSG) 0 bits`. We should be avoiding long words in pseudocode.

src/sha/sections/04-testtypes.adoc
```diff
@@ -36,36 +36,23 @@ For j = 0 to 99
...
-            Else:
-                C = TruncateToSize(MD, LEN(SEED))
-        Output MD
-        SEED = C
+For j = 0 to 99
+    A = B = C = SEED
+    For i = 0 to 999
+        MSG = A || B || C
+        if LEN(MSG) >= LEN(SEED):
+            MSG = TruncateToSize(MSG, LEN(SEED))
```
celic, line 47: Should just be `leftmost LEN(SEED) bits of MSG`

src/sha/sections/04-testtypes.adoc
livebe01: where?

jbrock24: >=, if it's equal we don't don't need to run the method it's going to call.

celic: `LEN(MSG) >= LEN(SEED)`

src/sha/sections/04-testtypes.adoc
```diff
@@ -36,36 +36,23 @@ For j = 0 to 99
...
-        Output MD
-        SEED = C
+For j = 0 to 99
+    A = B = C = SEED
+    For i = 0 to 999
+        MSG = A || B || C
+        if LEN(MSG) >= LEN(SEED):
+            MSG = leftmost LEN(SEED) bits of MSG
+        else:
+            MSG = MSG || LEN(SEED) - LEN(MSG) 0 bits
```
iDarshan, line 49: Concatenate MSG with LEN??

src/sha/sections/04-testtypes.adoc
iDarshan, line 46: Isn't it obvious that LEN(MSG) will always be greater than LEN(SEED)? Why this condition is needed?

src/sha/sections/04-testtypes.adoc
jbrock24: With 0's for bits, the LEN(SEED)-LEN(MSG). Padding 0's. We'll look into making this more clear, thank you.

src/sha/sections/04-testtypes.adoc
jbrock24: The MD returned from the SHA function will actually not be the LEN(SEED) but rather whatever the digest size is of the SHA you're running, i.e., 224/256.

Let's say given the sample of a SEED that is 1720 bits, the first iteration the MSG will be LEN(SEED)*3  or A = B = C = SEED. The next iteration it will be 1720 + 1720 + 256. The next iteration is 1720 + 256 + 256, the third is 256 + 256 + 256. So in this instance, we'll actually hit the condition in the third iteration of the inner loop.

PR #1462, opened 2023-08-22
Updated DRBG to show support for SHA3.

jbrock24:


livebe01:
This section should be updated as well: https://pages.nist.gov/ACVP/draft-vassilev-acvp-drbg.html#name-supported-values-per-drbg-o

jbrock24:
I think that's the only thing I changed, did you mean somewhere else? 

[src/drbg/sections/05-capabilities.adoc](https://github.com/usnistgov/ACVP/blob/b699ad27903d5a89b247691637f8d886a91c4284/src/drbg/sections/05-capabilities.adoc?plain=1#L56-L96)

jvdsn:
Is there any reference for SHA-3 DRBGs being FIPS approved? Or is this only added to ACVP to show support (similar to AES-GCM-SIV)?

jbrock24:
Hi @jvdsn, yes it's approved in a future version of 90A, they are good to use now, just takes a while to get the standard updated.

PR #1478, opened 2023-11-29
Made fixedPubExp an array in the sample registration. 

jbrock24:
I had updated the docs before changing it to array based on feedback.


abkarcher:
Quick question, in all other RSA related algorithms fixedPubExp is a string as opposed to an array. Is there a specific reason it is an array for this revision?

Thanks!
Andrew

jbrock24:
We're basically going to be broadening the registrations for some algorithms where this could be a possibility. IUT's are supporting more than 1 type of singular setting, and we want users to be able to submit a single registration for all their needs where possible. This is sort of the beginning. Thanks for the question!

PR #1500, opened 2024-04-01
Update 06-eddsa-siggen-test-vectors.adoc

jvdsn:
Cf. https://github.com/usnistgov/ACVP-Server/blob/master/gen-val/src/generation/src/NIST.CVP.ACVTS.Libraries.Generation/EDDSA/v1_0/SigGen/TestCase.cs#L22

### DIFF
--- a/src/eddsa/sections/06-eddsa-siggen-test-vectors.adoc
+++ b/src/eddsa/sections/06-eddsa-siggen-test-vectors.adoc
@@ -30,7 +30,7 @@ Each test group contains an array of one or more test cases. Each test case is a
 | tcId | Numeric identifier for the test case, unique across the entire vector set | integer
 | message | The message used to generate the signature | hex
 | context | The context string defined in FIPS 186-5 sections 7.6 and 7.8 | hex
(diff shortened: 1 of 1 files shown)


jvdsn:
Actually, I am wondering if contextLength should not also be part of the capabilities table at https://github.com/usnistgov/ACVP/blob/master/src/eddsa/sections/05-eddsa-siggen-capabilities.adoc ?

jvdsn:
According to FIPS 186-5 and https://github.com/usnistgov/ACVP-Server/blob/master/gen-val/src/orleans/src/NIST.CVP.ACVTS.Libraries.Orleans.Grains/Eddsa/OracleObserverEddsaDeferredSignatureCaseGrain.cs#L54, this is in octets (bytes), not bits.

jvdsn:
> Actually, I am wondering if contextLength should not also be part of the capabilities table at https://github.com/usnistgov/ACVP/blob/master/src/eddsa/sections/05-eddsa-siggen-capabilities.adoc ?

Based on my testing, this parameter is also provided as part of the request, so made some commits to add it to the table and the example registration.

jbrock24:
Thank you for the help @jvdsn!

PR #1597, opened 2025-07-09
Fix typos in the Ascon AEAD128 capabilities section

Kneckter:
The capabilities section for Ascon AEAD128 had two typos that do not follow the naming that the ACVP Server is looking for when registering for vectors. The reference JSON is here: https://github.com/usnistgov/ACVP-Server/blob/master/gen-val/json-files/Ascon-AEAD128-SP800-232/registration.json.

Note that the adLen was updated to aadLen because the server is looking for that key. AD means Associated Data and AAD means Additional Authenticated Data. I am not sure if the server should be changed to look for AD instead of AAD details, but the references I have seen in both projects use AAD for Ascon and call it Associated Data.

The details in `Test Case JSON Schema` do not have these typos.

### DIFF
--- a/src/ascon/sections/05-ascon-aead128-capabilities.adoc
+++ b/src/ascon/sections/05-ascon-aead128-capabilities.adoc
@@ -12,9 +12,9 @@ Each Ascon AEAD128 algorithm capability advertised is a self-contained JSON obje
 | algorithm | The algorithm to be validated | string | "Ascon"
 | mode | The Ascon mode to be validated | string | "AEAD128"
 | revision | The algorithm testing revision to use | string | "SP800-232"
-| directions | The directions for AEAD to be tested | array of strings | "encrypt", "decrypt"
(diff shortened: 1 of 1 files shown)


rossnist:
Hello, your pull request made us decide to stick with using "ad" rather than "aad". I created a separate pull request for our documentation which you can view here: https://github.com/usnistgov/ACVP/pull/1598
We have also updated our GenVals to reflect these changes.

Kneckter:
@rossnist, thanks for the review. Because of that decision, NIST will also need to change the [ACVP Server](https://github.com/usnistgov/ACVP-Server) as it does not use "adLen" or "ad". Is there a pull request or plan to make that change? I updated this repo the reflect what was accepted by the ACVP Server. With #1598, the two projects are out of sync.

rossnist:
I have a hotfix for the GenVals that is ready to go and should be reviewed/released within the next couple of days.

